PR #11, opened 2018-03-07
Added a raw working REPL web page as an extras

puria:


jaromil:
Awesome :^) soon we'll have an online demo this way

PR #18, opened 2018-04-18
linux shared lib targets

chespinoza:
I have been using these targets in the makefile to get a shared library in linux, it adds a pedantic way to tag the library ( ARCH VERSION BRANCH and GIT HASH  ) to track zenroom changes, but it is being pointed by a symlin called `libzenroom.so` in our machines.
I would like to know if this is ok to be merge or you think it needs some refinement.


jaromil:
utACK! great. you can ignore the small CI failure on JS tests, being fixed on develop already.
Will look into merging this next week.

PR #21, opened 2018-06-26
Feature/aead ecdh

ezimuel:
This PR adds the support of AES-[GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode) encryption/decryption in ecdh. It provides the following functions:

```lua
keyring:aead_encrypt(key, message, iv, header)
keyring:aead_decrypt(key, ciphertext, iv, header, tag)
```
where `header` is additional data and `tag` is the authentication tag generated by GCM encryption mode.

I also added the test vectors for AES-GCM by [IEEE P802.1](http://www.ieee802.org/1/files/public/docs2011/bn-randall-test-vectors-0511-v1.pdf)


ezimuel:
@jaromil on Travis there's a segmentation fault on [this test](https://travis-ci.org/DECODEproject/zenroom/jobs/396800149#L1198) for `osx`. I don't know why, thoughts?

src/zen_ecdh.c
```diff
@@ -352,9 +352,38 @@ static int ecdh_encrypt(lua_State *L) {
...
+	HERE();
+	ecdh *e = ecdh_arg(L, 1); SAFE(e);
+	octet *k = o_arg(L, 2); SAFE(k);
+	octet *in = o_arg(L, 3); SAFE(in);
+	octet *iv = o_arg(L, 4); SAFE(iv);
+	octet *h = o_arg(L, 5); SAFE(h);
+
+	// output is padded to next word
+	octet *out = o_new(L, in->len+16); SAFE(out);
+	octet *t = o_new(L, 16); SAFE (t);
```
jaromil: is this size 16 including a small boundary?
Elsewhere in Zenroom I use 2 bytes.
Padding may it may solve the current segfault in OSX.

ezimuel: Usually the tag size is 16 bytes. I can try adding a small boundary as you suggested.

jaromil: it wasn't the boundary in this case. will merge as tests are passing.

jaromil:
thanks for this much needed contribution :^D

PR #26, opened 2018-07-31
ElGamal with ZKP

kozko2001:
Added some Zero Knowledge Proof to elgamal.lua

- Verifies that the initial state is 0
- Verifies that the users only adds 0 or 1


jaromil:
awesome!

PR #28, opened 2018-08-09
Secretsharing

jaromil:
Fun implementation of sss by @dsprenkels inside Zenroom. Limited to 64B fixed size messages for now, will see how to move forward to variable size, meanwhile would love to know your opinion Daan!


jaromil:
On hold since no clear need for it has raised from pilots so far. While it may serve well some purposes in the future (threshold signature / decryption) it must be noted that the Coconut scheme already makes "internal" use of shamir secret sharing, implemented directly in Zenroom/Lua and making use of Milagro.

dsprenkels:
Awesome. :D

Maybe of some help: I do have a draft implementation of a variable-length implementation here: https://github.com/dsprenkels/sss/commit/bffd2afded3abae377b180ddc3efb97e6b37b3d1

I think you can use that as a prototype. Note, variable-length here means variable-length up to 4096 bytes. I'm keeping the requirement of being able to execute the code on platforms without any allocator.

In any case, I would recommend to implement any variable-length secret sharing in a memory safe language (which in your case would be Lua) if you can. Here's an [example](https://github.com/dsprenkels/sss-cli/blob/cec0a2e1c1d9b8af3f6f5a0fe23d65ce614f8a09/src/bin/split.rs#L99-L121) in Rust.

jaromil:
Thanks for the pointer Daan! yes indeed I'll be considering Lua.

jaromil:
Reference different implementation in milagro, issue #58

jaromil:
implemented in lua as lagrange interpolation

PR #30, opened 2018-08-24
ADD support for react-native

kozko2001:
Also fix non generating the `zenroom.js.mem` on newer emscripten sdk.

For make it work with react-native you also need to provide `crypto.js` file in the same folder and install buffer on the react-native app `npm install buffer`

```

var Buffer = require('buffer/').Buffer

const getRandomValues = (byteArray) => {
    for (let i = 0; i < byteArray.length; i++) {
        byteArray[i] = Math.floor(256 * Math.random());
    }
}
const randomBytes = (size, cb) => {
    if (size > 65536) throw new Error('requested too many random bytes')
    var rawBytes = new global.Uint8Array(size)

    // This will not work in older browsers.
    // See https://developer.mozilla.org/en-US/docs/Web/API/window.crypto.getRandomValues
    if (size > 0) {  // getRandomValues fails on IE if size == 0
        getRandomValues(rawBytes)
    }
    var bytes = Buffer.from(rawBytes.buffer)

    if (cb) {
        cb(bytes);
    }
    return bytes
}
module.exports = { randomBytes };
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,14 +46,29 @@ apply-patches:
 
 # TODO: improve flags according to
 # https://github.com/kripken/emscripten/blob/master/src/settings.js
-javascript-node: cflags += -DARCH_JS -D'ARCH=\"JS\"' --memory-init-file 1
-javascript-node: ldflags += --memory-init-file 1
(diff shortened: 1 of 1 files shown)


jaromil:
awesome!

PR #36, opened 2018-08-30
Modified elGamal to allow different votes

kozko2001:


jaromil:
awesome! `DATA` limit now fixed also in 99aa93c768b77d4f670c74d3539e3f0648105274.

PR #55, opened 2019-07-24
Tidy up javascript build targets

dappiu:
I think it makes more sense naming the javascript-* make rules after the binary format they emit, rather than after the environment for which they were created.
Emscripten by default emits code capable of running in Node, Browser, Shell, Web Worker, and this is also our case, so I replaced `-nodejs` with `-asmjs` (that was particularly misleading as recent Node versions could understand Wasm binaries too) and `-npm` with `-wasm`.

I also added a sample target for a browser-optimized Wasm binary (`emcc -S ENVIRONMENT='web'`) that's named after the output format first, and the specific environment after, so `javascript-wasm-web`, showing differences between output format and target environment.

### DIFF
--- a/build/javascript.mk
+++ b/build/javascript.mk
@@ -28,7 +28,8 @@ javascript-wasm: ldflags += -s WASM=1 \
 	-s MODULARIZE=1 \
 	-s ALLOW_MEMORY_GROWTH=1 \
 	-s WARN_UNALIGNED=1 \
-	-s EXPORT_NAME="'ZR'"
+	-s EXPORT_NAME="'ZR'" \
+	--no-heap-copy
 javascript-wasm: apply-patches lua53 milagro embed-lua
 	CC=${gcc} CFLAGS="${cflags}" LDFLAGS="${ldflags}" LDADD="${ldadd}" \
 	make -C src js
--- a/build/javascript.mk
+++ b/build/javascript.mk
@@ -39,6 +39,7 @@ javascript-wasm: apply-patches lua53 milagro embed-lua
 
 javascript-rn: cflags += -DARCH_JS -D'ARCH=\"JS\"' -D MAX_STRING=128000
 javascript-rn: ldflags += -s WASM=0 \
+	-s ENVIRONMENT=\"'shell'\" \
 	-s MODULARIZE=1 \
 	-s LEGACY_VM_SUPPORT=1 \
 	-s ASSERTIONS=1 \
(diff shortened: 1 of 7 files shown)


build/javascript.mk
```diff
@@ -2,7 +2,7 @@
 # https://github.com/kripken/emscripten/blob/master/src/settings.js
 javascript-node: cflags += -DARCH_JS -D'ARCH=\"JS\"' -D MAX_STRING=128000
 javascript-node: ldflags += -s WASM=0 \
-	--memory-init-file 1
+	-s MEM_INIT_METHOD=1
```
puria: is this https://github.com/emscripten-core/emscripten/issues/5108 okay already?

build/javascript.mk
```diff
@@ -32,10 +30,7 @@ javascript-rn: apply-patches lua53 milagro embed-lua
 	@cp -v src/zenroom.js 	  build/rnjs/
 
 javascript-demo: cflags  += -DARCH_WASM -D'ARCH=\"WASM\"' -D MAX_STRING=128000
-javascript-demo: ldflags += -s WASM=1 \
-	-s ASSERTIONS=1 \
-	-s NO_EXIT_RUNTIME=1 \
-	--shell-file ${extras}/shell_minimal.html
+javascript-demo: ldflags += -s WASM=1 --shell-file ${extras}/shell_minimal.html
```
puria: ask you to leave the ASSERTION=1 for debug purposes. Specially for the demo, that print stuff on the console.err if something goes wrong ;)

build/javascript.mk
```diff
@@ -39,6 +39,7 @@ javascript-wasm: apply-patches lua53 milagro embed-lua
 
 javascript-rn: cflags += -DARCH_JS -D'ARCH=\"JS\"' -D MAX_STRING=128000
 javascript-rn: ldflags += -s WASM=0 \
+	-s ENVIRONMENT=\"'shell'\" \
```
puria: I really don't have a way to test this... no project made with react-native by myself... but ok on trust and will check it soon over the decode app

build/javascript.mk
dappiu: Didn't see that issue before, but I can confirm that with emcc version 1.38.39, switching between `--memory-init-file 1` and `-s MEM_INIT_METHOD=1` produces identical results (binary identical .js and .js.mem memory initialization file).
But I also see that using `-s MEM_INIT_METHOD=0` for asm.js builds don't work as expected (do work on Wasm builds though), unlike `--memory-init-file 0` that do. So I'll change back to using the latter as it seems more reliable.

build/javascript.mk
dappiu: Good point, I missed the fact that some of the defaults read in https://github.com/emscripten-core/emscripten/blob/master/src/settings.js do change when optimization level is above 0, and `ASSERTIONS` is one of that. I'm going to put it back.

PR #56, opened 2019-07-26
Fix for #55

dappiu:
@puria


jaromil:
is this why the online demo is now broken?

PR #59, opened 2019-09-02
Fix some warnings and change extension of lua readme

claucece:
There are additional other warning that should be fixed ;)


jaromil:
thanks!

PR #75, opened 2019-10-21
Start checking some ecdh implementation

claucece:
This is still WIP. 

There are some interesting findings:
* Tests fails if an empty or NULL string is passed.
* The public key is defined as the double of its value. If corrected, the tests fail.

I will continue checking this ;)


claucece:
Ok , I finished the revision of the file. 

There is some lengths to be checked and some todos I've added ;)

@jaromil should we have a call to solve them all?

Thanks!

jaromil:
awesome! thanks

PR #80, opened 2019-11-29
Solve TODOs

claucece:
This checks:

* The length of some params (the iv and tag)
* Add references to RFCs.


jaromil:
many thanks!

PR #86, opened 2020-02-26
🐛 Fix the output concat double print in EMSCRIPTEN (thanks @0x01)

puria:


jaromil:
Great! the problem is fixed. This fix introduces a side-effect for Lua (but not for Zenroom) which prevents print() to accept multiple arguments in JS: only the last argument will be printed. However we use concat everywhere... just deviating from a standard in Lua. May be fixed later.

PR #87, opened 2020-02-26
Fix double print js problem also for write and error

puria:


puria:
Same as https://github.com/DECODEproject/Zenroom/pull/86#issuecomment-591430820 I do merge it for the new release over npm

PR #118, opened 2020-08-25
Remove hardcoded "ar" path in Makefile

zzappie:
Tiny fix removing hard-coded path that may cause make process to crash on some distros.


zzappie:
oops wrong branch

PR #119, opened 2020-08-25
Remove hardcoded ar path

zzappie:
Tiny fix removing hardcoded path to "ar" executable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,7 +117,7 @@ milagro:
 		cd ${pwd}/lib/milagro-crypto-c && \
 		CC=${gcc} LD=${ld} \
 		cmake . -DCMAKE_C_FLAGS="${cflags}" -DCMAKE_SYSTEM_NAME="${system}" \
-		-DCMAKE_AR=/usr/bin/ar -DCMAKE_C_COMPILER=${gcc} ${milagro_cmake_flags}; \
+		-DCMAKE_AR=${ar} -DCMAKE_C_COMPILER=${gcc} ${milagro_cmake_flags}; \
 	fi
 	if ! [ -r ${pwd}/lib/milagro-crypto-c/lib/libamcl_core.a ]; then \
 		CC=${gcc} CFLAGS="${cflags}" AR=${ar} RANLIB=${ranlib} LD=${ld} \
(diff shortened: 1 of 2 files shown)


zzappie:
Looking at travis logs... Can't figure out what is wrong...

The ar shell variable seems to be set adequately in configure.mk
But I find this .travis.yml line confusing:
```
- if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make win     && cp -v src/zenroom.exe build/ ; fi
- if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then make win-dll && cp -v src/zenroom.dll build/ ; fi
```
Is it a typo?

jaromil:
I have hardcoded that after many tries, long time ago. Not sure how it can be made better, but unless all tests pass I recommend not changing: by now there are many targets built by our CI and they could be affected.

zzappie:
I think I got what was going on.

Cmake has a quirk that is hard to spot.  During cmake run it generates
commands and stores them in files to later run them.  In our case the
problem was in file `CMakeFiles/amcl_core.dir/link.txt`.  If you don't
specify the full path to `ar` it asumes that `ar` binary sits in project
root.  So `link.txt` file eds up looking like this:

```
/home/user/path/to/zenroom/lib/milagro-crypto-c/ar qc lib/libamcl_core.a  CMakeFiles/amcl_core.dir/src/hash.c.o CMakeFiles/amcl_core.dir/src/rand.c.o CMakeFiles/amcl_core.dir/src/randapi.c.o CMakeFiles/amcl_core.dir/src/aes.c.o CMakeFiles/amcl_core.dir/src/gcm.c.o CMakeFiles/amcl_core.dir/src/oct.c.o CMakeFiles/amcl_core.dir/src/utils.c.o CMakeFiles/amcl_core.dir/src/version.c.o CMakeFiles/amcl_core.dir/src/pbc_support.c.o CMakeFiles/amcl_core.dir/src/ecdh_support.c.o
/gnu/store/h5v73fvcx4gf9j2xi6xr80icbfjgwkdc-profile/bin/ranlib lib/libamcl_core.a
```

But the interesting thing is that, because `ar` path was hardcoded,
`/usr/bin/ar` was used in crosscompiling milargo to windows with no
problems. :)

So to fix ci I just needed to specify full path to
`x86_64-w64-mingw32-ar`...

jaromil:
Many thanks for looking so deep into this! I've never been a fan of cmake and its quirks...

PR #120, opened 2020-08-25
Check that milagro makefile is generated before running clean target

zzappie:
Without this check `make clean` fails in case when milagro makefile hasn't been generated.


jaromil:
neat solution to an annoying little problem we had, thanks!

Issue #116, opened 2020-08-22
Messed up _zenroom.so file in PyPi package 1.0.12


zzappie:
Hello zenroom keepers!

I've been playing around with zenroom recently and stumbled upon an issue with with pypi package.

After trying to `from zenroom import zenroom` I got this traceback:

```
Traceback (most recent call last):
  File "/root/.local/lib/python3.6/site-packages/zenroom/zenroom_swig.py", line 26, in swig_import_helper
    pkg = __name__.rpartition('.')[0]
  File "/usr/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 994, in _gcd_import
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 953, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'zenroom._zenroom'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/root/.local/lib/python3.6/site-packages/zenroom/zenroom.py", line 5, in <module>
    from .zenroom_swig import zenroom_exec_tobuf, zencode_exec_tobuf
  File "/root/.local/lib/python3.6/site-packages/zenroom/zenroom_swig.py", line 32, in <module>
    _zenroom = swig_import_helper()
  File "/root/.local/lib/python3.6/site-packages/zenroom/zenroom_swig.py", line 26, in swig_import_helper
    pkg = __name__.rpartition('.')[0]
  File "/usr/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
ImportError: libpython3.7m.so.1.0: cannot open shared object file: No such file or directory
```

Which is strange since I have python 3.6.9 on that machine.
It turns out that not all _zenroom.so files are properly linked.

I went ran this ldd script in package directory.
```
for file in $(find | grep  .*_zenroom.so); do
    echo $(dirname $file) '->' $(ldd $file | grep -oE "libpython.{0,12}")
done
```
and got this:
```
./libs/Linux/3_6_2 -> libpython3.6m.so.1.0 libpython3.6m.so.1.0
./libs/Linux/3_6_5 -> libpython3.6m.so.1.0 libpython3.6m.so.1.0
[...]

PR #125, opened 2020-09-22
[WIP] Python bindings redesign

zzappie:
Hello Zenroom!

After opening issue #116 I've tried to modify current python bindings linking process, but after playing a bit more with current bindings implementation I stumbled upon various issues that in were harder to resolve than rewrite bindings.
 
First of all sorry for bluntly opening a pull request that tears of the part of the project :)  I wrote it for my project thought that with some adaptation you might find it useful too!

some problems with current implementation:
- tests are not passing
  I suspect the problem lies in tests, since I couldn't successfully run some test scenarios in apiroom (see `test_old.py`).
- calls to zenroom are made in separate processes.
   This prevents segfaults but often leaves Pool running. Deadlock? Something to do with file descriptors? Couldn't figure it out.
   Debugging this is hard.
- packaging process is complicated, error prone and creates unnecessary duplication of libs.
  pipy package weights 20Mb+
- package only works with cpython.
  Which is unjustified.  We only need to call couple of functions form libzenroom and using ctypes instead of swig will make it portable to pypy and remove the runtime swig code.

How this pr is organized:
Two commits:
- one removes old bindings and Makefile targets and adds target that just copies libzenroom.so to zenroom module folder.
   I've omitted version name and commit hash in library name for simplicity but it is good Idea to put it back and configure the python package during make run (there is `_configure.py file` just for that). 
- second adds new python bindings implementation.

How new implementation differs.
- It uses `ctypes` to load `zenroom_exec` and `zencode_exec` functions.
- It doesn't create separate process for calling these functions.
- there is `_redirect.py` module with convenient context managers capturing stdout and stderr into io stream from dynamically loaded library. 
- I named functions a bit differently (`lua_exec` and `zencode_exec`) to be more explicit but I can make the interface backward compatible if you will consider this pr.
- There are two levels of abstraction (`lua_exec` + `zencode_exe` and `lua_call` and `zencode_call`) first just calls zenroom functions in case user wants process it manually, the second captures output and returns `ZenResult` object.

Room for improvement:
- Since I'm calling zenroom functions in main thread the whole thing crushes on incorrect input. :)
  I can either implement input checking or or use multiprocessing. Or maybe it is possible to use some sort of error checks within libzenroom I don't know of?  This might be the thing someone could guide me.
- Little testing
- Configure libzenroom.so name during make run
- Installation time compilation?
   Maybe this would be a bit of overkill to compile lua, milagro and zenroom during python package installation.  But if libzenroom is self-contained and relocatable it should work fine precompiled.
- Optionally use `subproccess` to call zenroom executable if it is in PATH (because we can. the interface won't be much different).
- Backward compatibility.  Is it used in a lot of projects?
- Windows? Is it suposed to run on windows? I call things like `fileno` and `dup` so does the current implementation.
- Docs docstring etc...


puria:
Uuuh great work :tada:
`Swig` was a decision made in the first run also for `java` and future bindings, but since now it happens that the native implementations are more efficient as I said in #116, so I'm very grateful for your PR.

The tests probably are outdated since in the last month the zencode language had a massive refactoring and nobody cared about update the python tests (and sincerly I would wait at least another couple of weeks or so before having a stable zencode DSL version).

Hence ctypes are super welcome.

The separate process was made because of the way swig was engineered, in case of broken scripts the python interpreter did exit and no more output was shown. And in some wsgi/asgi servers stopped the working pool. So I wish to maintain multiprocessing to not kill the main thread.

Backward compatibility: it's used in almost 20 project as far as I know (10 of those are ours) but as we progress with the semantic versioning I'd rather go for a 2.x.x so, is not an issue. 

`zencode_exec` and `zenroom_exec` (I know they are super confusing) but we decided to keep them consistent over all the bindings, but I saw that you aliased the calls so :+1: 

As soon as you remove the [WIP] tag I'll make some testing and :twisted_rightwards_arrows: it.

jaromil:
Great work! about thread wrapping: at this point of development Zenroom should be completely isolating itself from the host without the need to wrap it in a thread. I think your approach should work without problems, but you do need proper error collection from stderr.

zzappie:
Thanks for kind feedback  @puria ad @jaromil!
I'll experiment with error checking and multiprocessing and will report improvements here

zzappie:
Closing in favor of #136

PR #129, opened 2020-10-16
fixed statement at line 158 of zencode_given.lua

Abbettowski:
"Given my name is in ' ' " should now work in every cases.

### DIFF
--- a/src/lua/zencode_given.lua
+++ b/src/lua/zencode_given.lua
@@ -156,8 +156,10 @@ end)
 Given("am ''", function(name) Iam(name) end)
 
 Given("my name is in ''", function(name)
-		 assert(IN[name], "No name found inside "..name)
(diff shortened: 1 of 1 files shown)


src/lua/zencode_given.lua
```diff
@@ -156,8 +156,10 @@ end)
 Given("am ''", function(name) Iam(name) end)
 
 Given("my name is in ''", function(name)
-		 assert(IN[name], "No name found inside "..name)
-		 Iam(IN[name])
+       pick(name, 'string')
+       assert(TMP, "No name found inside "..name)
+       local user = TMP.name
```
jaromil, line 161: no need, can to directly `Iam(TMP.name)` the value is kept in a global variable

Abbettowski:
adjusting in new fork with other fixes

PR #133, opened 2020-11-07
Abbettowski

Abbettowski:
Schnorr signature is ready


jaromil:
awesome new building block, congrats!

PR #136, opened 2020-11-20
Update python3 bindings

DanySpin97:
Supersedes #125.


puria (changes requested):
you want to remap lua_exec to `zenroom_exec`

zzappie:
I'm glad you have picked it up! @DanySpin97 
This pr was in my todo list for too long :)

zzappie:
Having both `..._exec`, `..._call` functions was design decision in case if user wants doesn't want to capture stdout and stderr.  This redirection stuff is very hacky.  You think `..._exec` are not necessary?

bindings/python3/zenroom/zenroom.py
```diff
@@ -33,55 +24,29 @@ def __post_init__(self):
...
-    _libzenroom_exec(_ZenLanguage.lua, script, conf, keys, data)
 
-
-def zencode_exec(script, conf=None, keys=None, data=None):
-    """Execute Zencode in Zenroom environment"""
-    _libzenroom_exec(_ZenLanguage.zencode, script, conf, keys, data)
-
-
-def _zenroom_call(language, script, conf, keys, data):
+def _apply_call(call, script, conf, keys, data):
```
zzappie, line 31: `_apply_call` -- seems to be strangely named to me. Apply means apply a function to arglist which essentially same as call a function.  Maybe just `_apply`?

zzappie: And also the `call` argument shouldn't it be `function`?
When I see `call(arg1, arg2, arg3)` I instantly think of `arg1(arg2, arg3)` or something like that.

zzappie:
BTW I see you haven't touched tests yet.
Pytest captures stdout so you'll need run it with `-s` flag tests won't work otherwise.

DanySpin97:
> Having both ..._exec, ..._call functions was design decision in case if user doesn't want to capture stdout and stderr. This redirection stuff is very hacky. You think ..._exec are not necessary?

The redirection stuff just replace stdout and stderr fds with one temporary file each. The shell uses the same method to make pipes work (`|`) so I'd say it's standard. I don't think it works in multi-threading, however multiple process writing on stdout isn't safe either (IMHO).

In the future there could be added API calls that works in multi-thread or directly fix `*_exec`, so having just these ones leaves the binding open for improvements.

bindings/python3/zenroom/zenroom.py
```diff
@@ -63,7 +63,7 @@ def zencode_exec(script, conf=None, keys=None, data=None):
 def _zenroom_call(language, script, conf, keys, data):
     outbuf = io.BytesIO()
     errbuf = io.BytesIO()
-    with output_to(outbuf), error_to(errbuf):
+    with redirect_sys_stream(True, outbuf), redirect_sys_stream(False, errbuf):
```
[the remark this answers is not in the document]
DanySpin97: > The purpose of two separate functions was exactly to get rid of obscure positional argument :)

This issue could be fixed by using a keyword argument. I think it's a better approach than having 2 (output_to and error_to) or 3 functions (output_to, error_to, common_to).

bindings/python3/zenroom/zenroom.py
DanySpin97: > _apply_call -- seems to be strangely named to me. Apply means apply a function to arglist which essentially same as call a function. Maybe just _apply?

I prefer a more verbose naming style, that's what I have used `_apply_call` instead of `_apply`:

```
def zenroom_exec(script, conf=None, keys=None, data=None):
    return _apply_call(_LIBZENROOM.zenroom_exec, script, conf, keys, data)
```

```
def zenroom_exec(script, conf=None, keys=None, data=None):
    return _apply(_LIBZENROOM.zenroom_exec, script, conf, keys, data)
```

> And also the call argument shouldn't it be function?

`call` is a `callable` hence the name, but it can be changed to function.

zzappie:
Danilo Spinella writes:

> The redirection stuff just replace stdout and stderr fds with one temporary file each. The shell uses the same method to make pipes work (`|`) so I'd say it's standard. I don't think it works in multi-threading, however multiple process writing on stdout isn't safe either (IMHO).
>
> In the future there could be added API calls that works in multi-thread or directly fix `*_exec`, so having just these ones leaves the binding open for improvements.

Okay I get your point.

zzappie:
Danilo Spinella writes:

> I prefer a more verbose naming style, that's what I have used
> `_apply_call` instead of `_apply`:

I is somewhat common idom in see here:
https://en.wikipedia.org/wiki/Apply

Some use `apply` some use `call` but apply_call raises questions :)

zzappie:
Danilo Spinella writes:
> This issue could be fixed by using a keyword argument. I think it's a
> better approach than having 2 (output_to and error_to) or 3 functions
> (output_to, error_to, common_to).

Dunno I like lots functions :)

zzappie:
Are you willing to take over the branch?  I may close my pr then.
Because my focus shifted away

jaromil:
@zzappie yes Zappie we are processing your contribution and will follow up on it, you will be credited as well for it, many thanks :^)

DanySpin97:
> Some use `apply` some use `call` but apply_call raises questions :)

Hmm right. Then `_apply_function` seems less ambigous.

DanySpin97:
> you want to remap lua_exec to zenroom_exec

To fix the tests? Is the design correct or is there anything else to change before working on the tests?

jaromil:
What is the status of this?

puria:
> What is the status of this?

@DanySpin97 is reworking on it... very close to finish ;)

puria (approved):

PR #141, opened 2020-12-11
changed 'vote' to 'preference'

Abbettowski:


jaromil:
Awesome! now able to process binary preferences

PR #147, opened 2021-01-13
fix: Add build directory for cmake

splasky:
The source code and the build code are mixed together. This
will produce some errors during compiling.

This commit adds a build directory for cmake.
To avoid mixing the source code and the build code together.

PR #148, opened 2021-01-15
fix: Fix errors when building the cortex-m target

splasky:
There are some compiling errors when building the cortex-m target.
Such as undefined reference to _sbrk(), _write(), _read()...etc,
use the deprecated API `zenroom_exec_rng_tobuf`, and the wrong brace
position.

This commit link the undefined reference to newlib nano specs, change
the deprecated API to zenroom_exec_tobuf and correct the brace position.


jserv:
#147 is somehow similar to this pull request. You should explain their difference.

PR #151, opened 2021-01-31
fix: Add semihosting support for Zenroom to run on qemu

splasky:
To use the semihosting will let us to test the Zenroom on
qemu-system-arm or such as cortex-m devices more easily.

But for running the semihosting correctly, we need to update the startup
file and the linker scripts.

This commit add a CMSIS repository into submodule for update the startup
file and the linker scripts to CMSIS and update the functions inside
cortex-m.c for semihosting support.

The Zenroom could be test on the following qemu scripts:
```
qemu-system-arm -M mps2-an385 -kernel src/zenroom.bin -semihosting -nographic
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "lib/cmsis"]
+	path = lib/cmsis
+	url = https://github.com/ARM-software/CMSIS_5.git
--- a/src/cortex_m.c
+++ b/src/cortex_m.c
@@ -1,250 +1,124 @@
 #include <stddef.h>
 #include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
 #include "zenroom.h"
+
 #ifdef ARCH_CORTEX
-extern unsigned int _start_heap;
-// #define NULL (((void *)0))
-//
-//
 
+#include <errno.h>
+#undef errno
+extern int errno;
 
-/**  XXX Temporary proxy until these dependencies are resolved **/
+enum semihost_ops
+{
+  SEMIHOSTING_SYS_OPEN = 0x01,
+  SEMIHOSTING_SYS_CLOSE = 0x02,
+  SEMIHOSTING_SYS_WRITEC = 0x03,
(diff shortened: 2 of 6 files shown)


jserv:
You should add some descriptions  about the usage of semihosting calls and your motivation in git commit messages.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "third_party/cmsis"]
+	path = third_party/cmsis
```
jserv, line 2: You can move the CMSIS to `lib/` directory for consistency.

.gitmodules
```diff
@@ -0,0 +1,3 @@
...
+	url = git@github.com:ARM-software/CMSIS_5.git
```
jserv, line 3: Use `https` for the protocol.

src/cortex_m.c
```diff
@@ -1,250 +1,124 @@
...
+  SYS_WRITE0 = 0x04,
+  SYS_WRITE = 0x05,
+  SYS_READ = 0x06,
+  SYS_READC = 0x07,
+  SYS_ISTTY = 0x09,
+  SYS_SEEK = 0x0A,
+  SYS_EXIT = 0x18,
+};
+
+static inline int semihosting_call(int R0, int R1)
```
jserv, line 28: Use shorter `semihost_call`.

src/cortex_m.c
```diff
@@ -1,250 +1,124 @@
...
-extern unsigned int _start_heap;
-// #define NULL (((void *)0))
-//
-//
 
+#include <errno.h>
+#undef errno
+extern int errno;
+
+enum semihosting_sys
```
jserv, line 14: Rename to `semihost_ops`.
See https://github.com/ContextualElectronics/stm32-bootloader/blob/master/bootloader/system/include/arm/semihosting.h

jserv (changes requested):
Apply the necessary changes against  src/cortex_m.ld. Don't copy file from CMSIS directly.

PR #153, opened 2021-02-08
ci: Validate Arm Cortex-M targets

splasky:
This commit adds the CI check for cortex-m targets.

The aps2-an386 needs the qemu version > 5.0. But the defulat
qemu version of the ubuntu's PPA is 4.2.0. So the target
changes from cortex-m4 to cortex-m3. When the ubuntu upgrade
the qemu version, the cflags can change back from cortex-m3
to cortex-m4.

For testing purpose, the RAM size is increase as large as possible.


jaromil (approved):
Very very good to have these tests running for Cortex-M targets.

The integration tests (test/zencode_*) aren't included yet, but we can take that as low priority: the whole point of integration tests is to simulate different moments of execution in a end-to-end sequence . I think we can rely on the crypto tests to establish this target is functioning, since they cover all mathematical primitives used in zencode.

PR #158, opened 2021-02-19
ci: Improve CI for cortex-m targets

splasky:
This commit adds the zencode integration test and the crypto
integration test for Arm cortex-m targets.

For testing the Zenroom on the cortex-m3 device, the stderr and stdout
stream should be changed. The stderr messages in Zenroom will be
redirect to semihosting console. The stdout message will be written
into the "stdout" file since the QEMU doesn't have the stderr and stdout
stream.

Since the QEMU also can not receive the stdin stream, the test script
will be written into a temp file. The QEMU will open the temp file
via semihosting to get the test scripts.

### DIFF
--- a/.github/workflows/cortex-arm.yml
+++ b/.github/workflows/cortex-arm.yml
@@ -4,13 +4,17 @@ on: [push, pull_request]
 
 jobs:
   coverity-scan:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
+      - uses: numworks/setup-arm-toolchain@2020-q4
       - name: checkout code
         uses: actions/checkout@v2
+      - name: update dependencies
+        run: |
+          sudo apt update && sudo apt upgrade
       - name: install dependencies
         run: |
-          sudo apt install gcc-arm-none-eabi zsh qemu-system-arm 
+          sudo apt install zsh qemu-system-arm
       - name: Checkout submodules
         uses: textbook/git-checkout-submodule-action@master
       - name: build cortex-arm target 
--- a/build/tests.mk
+++ b/build/tests.mk
@@ -64,6 +64,21 @@ cortex-m-crypto-tests = \
 	${1}test/coconut_test.lua && \
 	${1}test/coconut_abc_zeta.lua
 
+cortex-m-crypto-tests = \
+	test/cortex_m_crypto_tests.sh
+
+cortex-m-zencode-integration = \
+	cd test/zencode_numbers && ./run.sh ${1}; cd -; \
+	cd test/zencode_array && ./run.sh ${1}; cd -; \
+	cd test/zencode_hash && ./run.sh ${1}; cd -; \
+	cd test/zencode_ecdh && ./run.sh ${1}; cd -; \
+	cd test/zencode_credential && ./run.sh ${1}; cd -; \
+	cd test/zencode_petition && ./run.sh ${1}; cd -;
+
+cortex-m-crypto-integration = \
+	test/octet-json.sh ${1} && \
+	test/integration_asymmetric_crypto.sh ${1}
+	
 crypto-integration = \
 	test/octet-json.sh ${1} && \
 	cd test/nist && ./run.sh ../../${1}; cd -; \
@@ -238,10 +253,13 @@ check-crypto-debug:
 	$(call zencode-tests,${test-exec})
 	cat /tmp/zenroom-test-summary.txt
 
-check-cortex-m: test-exec := qemu-system-arm -M mps2-an385 -kernel src/zenroom.bin -semihosting -nographic
 check-cortex-m:
 	rm -f /tmp/zenroom-test-summary.txt
-	$(call cortex-m-crypto-tests,${test-exec} -semihosting-config arg=)
+	--rm -f ./outlog
+	$(call cortex-m-crypto-tests)
+	$(call cortex-m-crypto-integration,cortexm)
+	$(call cortex-m-zencode-integration,cortexm)
+	cat /tmp/zenroom-test-summary.txt
 	@echo "-----------------------"
 	@echo "All CRYPTO tests passed"
 	@echo "-----------------------"
--- a/test/cortex_m_crypto_tests.sh
+++ b/test/cortex_m_crypto_tests.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+
+[[ -r test ]] || {
+    print "Run from base directory: ./test/$0"
+    return 1
+}
+
+if ! test -r ./test/utils.sh; then
+    echo "run executable from its own directory: $0"; exit 1; fi
+. ./test/utils.sh
+
+run_zenroom_on_cortexm_qemu() {
+	qemu_zenroom_run "$*"
+	cat ./outlog
+}
+
+if [[ "$1" == "cortexm" ]]; then
+	zen=run_zenroom_on_cortexm_qemu
+fi
+
+crypto_tests=(
+	test/octet.lua
+	test/octet_conversion.lua
+	test/hash.lua
+	test/ecdh.lua
+	test/dh_session.lua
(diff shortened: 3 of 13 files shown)


splasky:
This PR should be merged after #153

jserv (changes requested):
Split the semihosting calls into another pull request(s), so that we can review the functional changes. The proposed CI pull request should address the integration issues though.

.github/workflows/cortex-arm.yml
```diff
@@ -4,15 +4,19 @@ on: [push, pull_request]
...
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
+      - uses: numworks/setup-arm-toolchain@2020-q4
       - name: checkout code
         uses: actions/checkout@v2
       - name: install dependencies
         run: |
-          sudo apt install gcc-arm-none-eabi zsh
+          sudo apt install zsh qemu-system-arm jq
```
jserv, line 14: `jq` would be regarded as the enhancements, which can be decoupled from CI work.

jserv: Undo it to minimize the necessary changes in single pull request.

build/tests.mk
```diff
@@ -49,6 +49,40 @@ crypto-tests = \
...
+	# cd test/zencode_given && ./run.sh; cd -; \
+	# ./test/zencode_parser.sh ${1} && \
+
+cortex-m-crypto-integration = \
+	test/octet-json.sh ${1} && \
+	test/integration_asymmetric_crypto.sh ${1}
+	# not pass yet
+	# cd test/nist && ./run.sh ../../${1}; cd -;\
+	
+
```
jserv, line 85: Remove this blank line for consistency.

test/octet-json.sh
```diff
@@ -4,11 +4,20 @@ tstr="Zenroom test"
 zenroom=$1
 valgrind=$2
 
+qemu_zenroom_run(){
+	qemu-system-arm -M mps2-an385 -kernel src/zenroom.bin -semihosting -nographic -semihosting-config arg="$*"
+	cat ./stdout
```
jserv, line 9: `stdout` might be confusing. You can rename it to `out` or `outlog` (logger of STDOUT).

build/tests.mk
```diff
@@ -49,6 +49,39 @@ crypto-tests = \
 	${1} test/coconut_test.lua && \
 	${1} test/crypto_abc_zeta.lua
 
+cortex-m-crypto-tests = \
+	${1}test/octet.lua && cat ./stdout &&\
```
jserv, line 53: You did not check if the file `stdout` was generated by previous command or something even earlier. That is, if LUA script was returning 0 without generating file `stdout`. The sequence would not be aware.

test/zencode_parser.sh
```diff
@@ -6,17 +6,29 @@ echo "============================="
 echo "ERRORS ARE OK IN THIS SECTION"
 
 alias zenroom="${1:-./src/zenroom}"
+
+qemu_zenroom_run(){
+	qemu-system-arm -M mps2-an385 -kernel src/zenroom.bin -semihosting -nographic -semihosting-config arg="$*"
```
jserv, line 11: Remove file `stdout` before executing `qemu-system-arm`.

build/tests.mk
```diff
@@ -222,6 +255,18 @@ check-crypto-debug:
 	$(call zencode-tests,${test-exec})
 	cat /tmp/zenroom-test-summary.txt
 
+check-cortex-m: test-exec := qemu-system-arm -M mps2-an385 -kernel src/zenroom.bin -semihosting -nographic
```
jserv, line 258: `qemu-system-arm` appears in various parts. You can split it into a dedicated script and then use `source`.

build/tests.mk
```diff
@@ -49,6 +49,39 @@ crypto-tests = \
...
+	rm ./outlog && ${1}test/coconut_abc_zeta.lua && cat ./outlog
+
+cortex-m-zencode-integration = \
+	cd test/zencode_numbers && ./run.sh ${1}; cd -; \
+	cd test/zencode_array && ./run.sh ${1}; cd -; \
+	cd test/zencode_hash && ./run.sh ${1}; cd -; \
+	cd test/zencode_ecdh && ./run.sh ${1}; cd -; \
+	cd test/zencode_credential && ./run.sh ${1}; cd -; \
+	cd test/zencode_petition && ./run.sh ${1}; cd -;
+	# not pass yet
```
jserv, line 74: Since there are remaining test items which would expectedly fail, you should mention in git commit message(s).

jserv: with explanation explicitly.

test/utils.sh
```diff
@@ -2,49 +2,53 @@
...
-				if test -r "$p/zenroom.exe"; then zenroom_path="$p/zenroom.exe"; break; fi
-			done
-		;;
-		bsd*)
-			for p in "${zenroom_paths[@]}"; do
-				if test -r "$p/zenroom"; then zenroom_path="$p/zenroom"; break; fi
-			done
-		;;
-	esac
+	if test "$is_cortexm" == true; then
```
jserv, line 5: If you move this `if` statement right after switch-esac statement, you can minimize the proposed changes with fewer lines.

build/tests.mk
jserv: Instead of invoking the qemu with predefined arguments, use a wrapper script for abstraction.

build/tests.mk
splasky: Some test files don't pass yet. It will be modified in another PR to be passed.

test/octet-json.sh
```diff
@@ -4,11 +4,33 @@ tstr="Zenroom test"
...
-	if [[ "$2" != "valgrind" ]]; then
-		$zenroom $*
-	else
+	if [[ "$valgrind" == "valgrind" ]]; then
 		valgrind --max-stackframe=2064480 $zenroom $*
+	else
+		if [[ "$zenroom" == "cortexm" ]]; then
+			run_zenroom_on_cortexm_qemu $*
+		else
+			$zenroom $*
```
jserv, line 32: If `zenroom` was properly overridden, then we can use the unified `zenroom` for execution.

test/zencode_petition/run.sh
```diff
@@ -9,6 +9,12 @@
 if ! test -r ../utils.sh; then
 	echo "run executable from its own directory: $0"; exit 1; fi
 . ../utils.sh
+
+is_cortexm=false
```
jserv, line 13: Can you eliminate (or reduce) the use of `is_cortexm`?

splasky: I think it is better to keep the "is_cortexm" variable. Since the "Z" is the function, the "$1" will be overridden when passing the argument.

test/cortex_m_crypto_tests.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+[[ -r test ]] || {
+    print "Run from base directory: ./test/$0"
+    return 1
+}
+
+if ! test -r ./test/utils.sh; then
+	echo "run executable from its own directory: $0"; exit 1; fi
```
jserv, line 9: Improve the indention.

test/utils.sh
```diff
@@ -71,6 +78,15 @@ detect_zenroom_conf() {
 	unset zenroom_conf
 }
 
+qemu_zenroom_run(){
```
jserv, line 81: Rewrite as `qemu_zenroom_run() {`

jaromil (approved):
It's overall good. We all know there are things that could be made more 'elegant', but this is not actual zenroom code but test shell scripts modified with workarounds to add the qemu execution, so we can be fine with this. Thanks!

splasky:
This should be merge after #159

PR #159, opened 2021-02-20
Implement print stream for cortex-m target

splasky:
This commit implements the output function for cortex-m target.
Since the QEMU didn't have the stderr and stdout stream. The
cortex-m device need to write the stderr message to semihosting
console via semihosting call. The messages of stdout will be written
into a "stdout" file.

The output will help CI to validate the cortex-m result in QEMU.

### DIFF
--- a/src/zen_io.c
+++ b/src/zen_io.c
@@ -39,14 +39,23 @@
 #define MAX_JSBUF 4096000 // 4MiB
 #endif
 #endif
+
+#if defined(ARCH_CORTEX)
+extern int SEMIHOSTING_STDOUT_FILENO;
+extern int write_to_console(const char* str);
+#endif
+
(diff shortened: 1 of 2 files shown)


splasky:
This PR should be merged after #153

src/zen_io.c
```diff
@@ -39,14 +39,23 @@
 #define MAX_JSBUF 4096000 // 4MiB
 #endif
 #endif
+
+#if defined(ARCH_CORTEX)
```
jserv, line 43: Create a pull request for revising the use of `ARCH_CORTEX`. Something like `CORTEXM` or `CORTEX_M`.

splasky: Need to discuss: #160

jaromil (approved):

PR #164, opened 2021-03-02
ci: Add check-crypto CI for all platform

splasky:
There are several platforms that Zenroom support. Some changes on other specific platform may not easily to check such as cortex m.

This commit adds a github actions CI to check Zenroom will pass on ubuntu, macos and windows platform.

### DIFF
--- a/.github/workflows/check-crypto.yml
+++ b/.github/workflows/check-crypto.yml
@@ -1,0 +1,24 @@
+name: Check crypto
+
+on: [push, pull_request]
+
+jobs:
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


puria (changes requested):
I'd rather prefer to have `crypto-check` here in place of platform specific, with a matrix.os including other platforms (mac, windows), what you think?

splasky:
> I'd rather prefer to have `crypto-check` here in place of platform specific, with a matrix.os including other platforms (mac, windows), what you think?

Great idea! Let me modify it.

splasky:
@jaromil The CI may fail to build on macOS. Can you check it?

splasky:
By the way, the travis CI will fail since the modified for src/cli.c inside commit 14e4103.

puria (approved):

jaromil:
many thanks!! this is a great achievement to prove functioning on cortex chips

PR #181, opened 2021-04-12
Do not add main function into zenroom library when using meson

DanySpin97:


DanySpin97:
Closed in favor of #185.

PR #183, opened 2021-04-14
[WIP] Update go bindings

DanySpin97:


bindings/golang/zenroom/zenroom.go
```diff
@@ -3,86 +3,29 @@
...
-// verbose Zenroom should be with its output. The value of this configuration
-// must be an integer from 1 to 3, where 1 is the least verbose, and 3 is the
-// most. The default if this value is not supplied is 1.
-func WithVerbosity(verbosity int) Option {
-	return func(c *config) {
-		c.Verbosity = verbosity
-	}
+type ZenResult struct {
+    output string;
+    err string;
```
puria, line 28: Would prefer to rename it in logs to keep it consistent with js. Also actually are the output of the execution, not just errors... what you think?

puria (changes requested):

bindings/golang/zenroom/zenroom.go
DanySpin97: I agree with you about using `logs` over `err`. I have used the same naming as python as I didn't know that js used `logs`.

puria (approved):

PR #211, opened 2021-07-14
Add a Global Interpreter Lock to Rust binding

jpopesculian:
* version bump to v0.2.0
* adds Global Interpreter Lock
* makes API arguments / return types more "Rusty"
* adds Cargo.lock to .gitignore (see: https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html)


jaromil:
Thanks! looks pretty well done, only concern is about the added deps:
```rust
lazy_static = "1.4"
thiserror = "1.0"
```
how risky is this in terms of future compatibility, forced changes and relying on the security of other projects?

I am asking because so far we considered as a feature the complete lack of deps of Zenroom.

I do not understand well the functionalities brought by these two deps, so I better ask: could they be optional for a certain build flavor / binding branch and then still offer a dependency-free zenroom binding?

We can debate this also in a quick call when you have time

jpopesculian:
Hey Jaromil, sure, I understand the want to be dependency less! They were mainly convenience and best practice / readability things, but I reimplemented them without the dependencies. Let me know what you think

jaromil:
many thanks :^)

PR #248, opened 2021-09-15
Fix base58 encoding

albertolerda:


jaromil:
Well spotted!! did not realise maxlen was read/write

PR #251, opened 2021-09-25
fixes android build linking error

adam-burns:
This fixes Android builds by removing inherited link libraries (ie -lpthread) in ldflags option.


jaromil:
thanks!!

PR #253, opened 2021-09-30
BIP-39 encoding with tests

albertolerda:
I have imported the implementation of BIP-39 from https://github.com/trezor/trezor-crypto/blob/master/bip39.c with some changes to make it compatible with zenroom, all the modifications are described in the file encoding.c .
Official test vectors have been imported.
The standard support the encoding of octets of length at most 256 bits


jaromil:
another amazing contribution, well done!

PR #256, opened 2021-10-04
Fixed prime in lagrange interpolation

albertolerda:
The new prime is 32bytes long, for this version to work we need
at least 257bits big integer (the sum of two 256bits integer could
be a 257bits integer). Currently we are using 384bits big integer and
they are fine.


jaromil:
very well done, now secret sharing works with any 32byte secret

PR #260, opened 2021-11-03
Prefix milagro exports which conflict with OpenSSL

jpopesculian:


jaromil:
Very cool to fix this for any use of libzenroom! thanks

if you like, you are welcome to submit this patch also to 
https://github.com/dyne/lua-zenroom 

else I will port it there and acknowledge your contrib

jpopesculian:
Ah cool! I didn't know about lua-zenroom! I can take a look

PR #276, opened 2021-12-14
Forse line endings in unix eol's

puria:


jaromil:
seems the gitattributes needs license attribution, I suggest public domain?
```
The following files have no copyright and licensing information:
* .gitattributes
```

puria:
> seems the gitattributes needs license attribution, I suggest public domain?

NO, it's our awesome ReUse license check! Remember we are https://github.com/fsfe/reuse-tool compliant so every file (like this new one) need an indication of the licensing ;)

PR #280, opened 2022-01-02
Update rust bindings

danyspin97:
Using `String` as function argument, means that the variable will be moved and thus consumed when the function is invoked. This might be good in some cases, where the string passed isn't needed anymore. However, in other cases a new allocation is required to avoid consuming the variable. Since a `CString` (which is just a wrapper over `char*`  is allocated in the bindings anyway, passing `&str` is better way. Also, `&str` is more flexible and thus preferred to `String` as function argument.

This PR also contains an update to the dependencies.

### DIFF
--- a/bindings/rust/src/lib.rs
+++ b/bindings/rust/src/lib.rs
@@ -68,29 +68,29 @@ type Fun = unsafe extern "C" fn(
 ) -> ::std::os::raw::c_int;
 
 pub fn zencode_exec(
-    script: &str,
-    conf: &str,
-    keys: &str,
-    data: &str,
+    script: impl AsRef<str>,
+    conf: impl AsRef<str>,
+    keys: impl AsRef<str>,
+    data: impl AsRef<str>,
 ) -> Result<ZenResult, ZenError> {
     exec_f(c::zencode_exec_tobuf, script, conf, keys, data)
 }
 
 pub fn zenroom_exec(
-    script: &str,
-    conf: &str,
-    keys: &str,
-    data: &str,
+    script: impl AsRef<str>,
+    conf: impl AsRef<str>,
+    keys: impl AsRef<str>,
+    data: impl AsRef<str>,
 ) -> Result<ZenResult, ZenError> {
     exec_f(c::zenroom_exec_tobuf, script, conf, keys, data)
 }
 
 fn exec_f(
     fun: Fun,
-    script: &str,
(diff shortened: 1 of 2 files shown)


jaromil:
@jpopesculian can you share your review about this? would it be OK to merge? I think you are the one most affected by this chance.

jpopesculian (approved):
Awesome! A much easier calling interface

bindings/rust/src/lib.rs
```diff
@@ -68,29 +68,29 @@ type Fun = unsafe extern "C" fn(
 ) -> ::std::os::raw::c_int;
 
 pub fn zencode_exec(
-    script: String,
-    conf: String,
-    keys: String,
-    data: String,
+    script: &str,
```
jpopesculian, line 71: If you wanted to make the interface even easier you could do something like `script: impl AsRef<str>` and then you would do the same thing but wouldn't break backwards compatibility.

danyspin97: I have tried your suggestion but I couldn't make the code compile. Since I have never used `AsRef` I might be doing something wrong, can you please try and share the patch/commit? Thank you!

albertolerda: Try this snippet for the function `exec_f`, I think that it works!
```
fn exec_f(
    fun: Fun,
    script: impl AsRef<str>,
    conf: impl AsRef<str>,
    keys: impl AsRef<str>,
    data: impl AsRef<str>,
) -> Result<ZenResult, ZenError> {
    let mut stdout = Vec::<i8>::with_capacity(BUF_SIZE);
    let stdout_ptr = stdout.as_mut_ptr();
    let mut stderr = Vec::<i8>::with_capacity(BUF_SIZE);
    let stderr_ptr = stderr.as_mut_ptr();

    let lock = aquire_zen_gil();
    let exit_code = unsafe {
        fun(
            CString::new(script.as_ref())?.into_raw(),
            CString::new(conf.as_ref())?.into_raw(),
            CString::new(keys.as_ref())?.into_raw(),
            CString::new(data.as_ref())?.into_raw(),
            stdout_ptr,
            BUF_SIZE as u64,
            stderr_ptr,
            BUF_SIZE as u64,
        )
    };
    drop(lock);
```

jaromil:
awesome! thanks everyone

PR #313, opened 2022-03-08
Fixed test for Dilithium

matteocristino:


puria:
🥳 bomba!
You also want to add the licensing credits of `lib/pqclean` to https://github.com/dyne/Zenroom/blob/master/.reuse/dep5 (look at the file, should be trivial) to maintain the Reuse compliancy

PR #321, opened 2022-03-23
Schnorr signature: some improvement

matteocristino:
Improved Schnorr signature functions and added tests for the new functions


matteocristino:
I apologize for the mess and I thank you for the proposal. I will create a new PR in the next days from a different branch starting from the remote/master branch with the same commit so it will be more clean and none wates his time on this.

PR #327, opened 2022-03-28
Zencode: create flat array

matteocristino:
New Zencode statement to flat an array

### DIFF
--- a/src/lua/zencode_array.lua
+++ b/src/lua/zencode_array.lua
@@ -285,7 +285,6 @@ end)
 When("create the flat array of contents in ''", function(dic)
 	local codec = ZEN.CODEC[dic]
 	ZEN.assert(codec.zentype == 'array' or
-		   codec.zentype == 'schema' or
 		   codec.zentype == 'dictionary',
 		   "Target is not a valid: "..dic)
 	local data = have(dic)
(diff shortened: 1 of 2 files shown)


src/lua/zencode_array.lua
```diff
@@ -281,3 +281,26 @@ When("create the standard deviation of elements in array ''", function(arr)
 	ACK.standard_deviation = O.from_string(stats.standardDeviation(data):decimal())
 	new_codec('standard_deviation', { encoding="string" })
 end)
+
+local function _flat_array(data, res)
+   for _, item in ipairs(data) do
```
jaromil, line 286: should use `pairs()` instead of `ipairs()` here and in general best use `pairs` everywhere

jaromil:
To justify the introduction of a new noun `flat array` it would be best to make it applicable to all cases, also deep dictionaries. For that to happen I recommend adopting `deepmap()` passing it a function that does `table.insert`.

jaromil:
For completeness, perhaps two statements
```
When I create the flat array of contents in ''
When I create the flat array of keys in ''
```

jaromil:
well done! we reviewed this yesterday. I am wondering about the term "contents" and if it should be "values" but this is a minor change I can apply, meanwhile we'll be merging your contribution, well done!

PR #328, opened 2022-03-29
small code to parse the documentation

matteocristino:


puria (commented):
Good job ;) Maybe could be a good idea to have a Github Action that checks the documented statements and eventually shows a warning if not found... is that feaseable?

docs/doc_needed/run.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
+
+# IMPORTANT: the binary file of zenroom has to be in zenroom/src/
+
+# create the file needed to store the data
+>introspection.json
+>documented.txt
```
albertolerda, line 7: What do you mean with create? If you do `>>` and the file doesn't exist it is created

docs/doc_needed/doc_control.lua
```diff
@@ -0,0 +1,103 @@
...
+   local d = {}
+   local scenarios = {}
+   local scenario, line
+   
+   while true do
+      line = io.read()
+      if not line then break end
+      
+      line = string.gsub(line, "^%s*(.-)%s*$", "%1")
+      if line == "{" then
```
albertolerda, line 19: There is no certainty that the file json will be generated with a '{' on a single line... Is it sufficient to use a custom format like
```
scenario
statement
statement
statement

scenario
statement
....
```

docs/doc_needed/run.sh
```diff
@@ -45,15 +47,22 @@ echo "  all documented statements have been loaded   "
...
 echo "-----------------------------------------------"
 echo
 
-rm -f introspection.json documented.txt
+lines=`wc -l < to_be_documented.txt`
+
+if [[ $lines -gt $count ]]
+then
+    to_be_doc=$((lines-count))
+    echo "There are ${to_be_doc} statements to be documented"
```
albertolerda, line 66: What about showing the statements that need the documentation?

PR #345, opened 2022-04-11
fix: Schnorr zencode not signing arrays and dictionaries

matteo-cristino:


jaromil:
seems very good to me

PR #351, opened 2022-04-15
First version of Streamline NTRU Prime in lua

matteo-cristino:
For the moment all the checks (pk, ss, ct) are done looking only at the length. Now I'm looking deeper in the algo to see if there is the possibility of implementing some cryptographic check.


src/zen_qp.c
```diff
@@ -27,100 +27,99 @@
...
 extern int PQCLEAN_KYBER512_CLEAN_crypto_pub_gen(uint8_t *pk, uint8_t *sk);
-
 extern int PQCLEAN_KYBER512_CLEAN_crypto_kem_enc(uint8_t *ct, uint8_t *ss, const uint8_t *pk);
-
 extern int PQCLEAN_KYBER512_CLEAN_crypto_kem_dec(uint8_t *ss, const uint8_t *ct, const uint8_t *sk);
 
+/*
+  Quantum proof Streamlined NTRU Prime kem/cipher
+*/
+#define PQCLEAN_SNTRUP761_CLEAN_CRYPTO_SECRETKEYBYTES  1763
```
jaromil, line 61: it makes sense to use this define in the zencode implementation
which in turn calls for the need of a header for zen_qp.h, about time for it, so many externs...

src/zen_qp.c
```diff
@@ -202,76 +201,80 @@ static int qp_verify(lua_State *L) {
...
 static int qp_kem_keygen(lua_State *L) {
 	lua_createtable(L, 0, 2);
-        octet *private = o_new(L,PQCLEAN_KYBER512_CLEAN_CRYPTO_SECRETKEYBYTES); SAFE(private);
+	octet *private = o_new(L,PQCLEAN_KYBER512_CLEAN_CRYPTO_SECRETKEYBYTES); SAFE(private);
 	lua_setfield(L, -2, "private");
-        octet *public = o_new(L,PQCLEAN_KYBER512_CLEAN_CRYPTO_PUBLICKEYBYTES); SAFE(public);
+	octet *public = o_new(L,PQCLEAN_KYBER512_CLEAN_CRYPTO_PUBLICKEYBYTES); SAFE(public);
 	lua_setfield(L, -2, "public");
-
+
```
jaromil, line 226: whoeps the whitespaces

src/lua/zencode_qp.lua
```diff
@@ -118,104 +157,97 @@ ZEN.add_schema(
 		       export = O.to_hex },
       kyber_ciphertext = { import = kyber_ciphertext_f,
 			   export = O.to_hex },
-      kyber_kem = { import = import_kem,
-		    export = export_kem }
+      kyber_kem = { import = kyber_import_kem,
+		    export = kyber_export_kem },
+
+      
+      ntrup_public_key = { import = ntrup_public_key_f,
```
jaromil, line 164: the "complex" schema defining import/export functions can be simplified, just one function is enough and will use the encoding set by `rule` or default base64. this can be done since there is no established standard for encoding ntru keys, so any encoding is good. other cases use import/export where a specific standard is needed (i.e in bitcoin, eth...)

can explain this in a meeting soon

jaromil (approved):
overall good, can merge this and apply fixes mentioned later

PR #361, opened 2022-05-04
clean up and improved filter operation for dictionaries

matteo-cristino:


src/lua/zencode_dictionary.lua
```diff
@@ -272,27 +272,33 @@ When("move '' in ''", function(src, dict)
...
+      if fv:str() == k then keep = true end
+   end
+   if not keep then res = nil end
+   return res
+end
+
+local function _is_array_of_dictionaries(a)
+   if not isarray(a) then return false end
+   for _, v in pairs(a) do
+      if not isdictionary(v) then
```
jaromil, line 288: this operation is quite expensive in a loop, lets keep an eye on its impact

PR #364, opened 2022-05-06
now it is possible to branch on 'found'

matteo-cristino:


jaromil:
Let's elaborate a bit more on this case. Ideas to add:
1. `when/if '' is found in ''`
2. internal lua function `isfound()` could check for more than `not nil` but also empty octet (zero size)

PR #365, opened 2022-05-06
given nothing update

matteo-cristino:
In the given phase DATA and KEYS are always set to nil. Their content is moved in, respectively, IN and KIN, so we need to check the emptiness of the latters.


jaromil:
well spotted!

PR #373, opened 2022-05-25
fix: json parsing

matteo-cristino:


jaromil:
thanks!! the devil is in the details...

PR #390, opened 2022-06-08
Zencode ethereum nonce has to be a number

matteo-cristino:


albertolerda:
It's nice that in lua the only falsy values are `nil` and `false`! 🤩

PR #401, opened 2022-06-30
fix(w3c): typo in the verification method schema

matteo-cristino:


andrea-dintino:
1 line fix, we tried the fix together

PR #403, opened 2022-07-01
fix: random dictionary creation

matteo-cristino:


jaromil:
well done!

jaromil:
in light of https://github.com/dyne/Zenroom/commit/cdceedc7855915a358a234b8ce21219635c88dc9 and latest issue on passing a deterministic seed runtime I believe best approach here is to pass a pointer to the random function as argument to the local function.

matteo-cristino:
now the local function takes as argument also the pointer to the random function

Issue #420, opened 2022-07-21
some math operations are not working


matteo-cristino:
Using the following input:
```
{
	"mySecondNumber": 2,
	"myThirdNumber": 3
}
```
with the script:
 ```
Given I have a 'number' named 'mySecondNumber'
Given I have a 'number' named 'myThirdNumber'
When I create the result of 'mySecondNumber' % 'myThirdNumber'
and I rename the 'result' to 'resultOfmyFirstModulo'
Then print data 
```
The following error is reported:
[W] [!] /zencode_when.lua:315: attempt to perform arithmetic on a zenroom.float value (local 'l')

or with the script:
```
Given I have a 'number' named 'mySecondNumber'
Given I have a 'number' named 'myThirdNumber'
When I create the result of 'mySecondNumber' inverted sign
and I rename the 'result' to 'resultOfmyFirstModulo'
Then print data
```
The following error is reported:
[W] [!] /zencode_when.lua:353: Incompatible numeric arguments

Moreover these statements are present in run-when.sh file (https://github.com/dyne/Zenroom/blob/master/test/zencode_cookbook/run-when.sh) which is executed in the actions, but during them no error is reported.

Issue #419, opened 2022-07-21
"set ' ' to ' ' as ' ' " does not work with numbers


matteo-cristino:
Running the following script:
```
Given nothing
When I set 'myNewlytCreatedNumber' to '42' as 'number'
Then print data
```
the following error is reported:
[!] /zencode_when.lua:133: bad argument #1 to 'from_decimal' (string expected)

PR #421, opened 2022-07-21
Fix numbers (fixing #420 and #419)

albertolerda:


jaromil:
well done! merged fast as I need isnumber() already :^)

PR #423, opened 2022-07-25
fix(array): split at spaces

matteo-cristino:


jaromil:
once again well spotted

PR #428, opened 2022-07-27
New zencode statement for kdf and pbkdf

matteo-cristino:


jaromil:
well done! future documentation should mention default number of rounds and sane settings (we will test on mobile and average PCs using apiroom)

Issue #431, opened 2022-08-02
When '' is found '' (number/float issue)


andrea-dintino:
Script: 

```
Given I have a 'string array' named 'haystack'
Given I have a 'number' named 'quorum'
Given I have a 'string' named 'needle'

When the 'needle' is found in 'haystack' at least 'quorum' times

Then print data
```

data

```
{
	"myArrayToBeHashed": [
		"john@beatles.com",
		"paul@beatles.com",
		"ringo@beatles.com",
		"george@beatles.com"
	],
	"myString": "I love Liverpool!",
	"string2": "Approved",
	"haystack": [
		"Approved",
		"Not approved",
		"Approved",
		"junk data blah blah 123456",
		"Approved",
		12345678,
		"Approved",
		{
			"stuff": "blah blah blah"
		},
		"Approved",
		[
			"some more junk"
		]
	],
	"quorum": 5,
	"needle": "Approved"
}

```

Output: 

```
 .  Build commit hash: 94ff505c
 .  ECDH curve is SECP256K1
 .  ECP curve is BLS381
 .  Memory in use: 576 KB
[*] reading Zencode from file: needle.zen
 .  loaded file (217 bytes)
[*] Direct Zencode execution
[W] Zencode is missing version check, please add: rule check version N.N.N
[W] When the 'needle' is found in 'haystack' at least 'quorum' times    2
[W] nil
[W]  D  ZEN:pick found haystack(array)
[W]  D  ZEN:pick found quorum(element)
[W]  D  ZEN:pick found needle(element)
[W] [!] [string "zencode_array"]:249: bad argument #2 to '__eq' (float class exp                                                                                                       ected)
[W] {
    a_GIVEN_in = {},
    b_GIVEN_in = {},
    c_WHEN_ack = {
        haystack = [10] { octet[8] 417070726f766564, octet[12] 4e6f7420617070726                                                                                                       f766564, octet[8] 417070726f766564, octet[26] 6a756e6b206461746120626c616820626c                                                                                                       616820313233343536, octet[8] 417070726f766564, float 1.234568e+07, octet[8] 4170                                                                                                       70726f766564, {
                stuff = octet[14] 626c616820626c616820626c6168
            }, octet[8] 417070726f766564, [1] { octet[14] 736f6d65206d6f7265206a                                                                                                       756e6b } },
        needle = octet[8] 417070726f766564,
        quorum = float 5
    },
    d_THEN_out = {}
}
[!] ERROR:
[...]

PR #451, opened 2022-08-10
only compare if same type (fixing #431)

albertolerda:


src/lua/zencode_array.lua
```diff
@@ -246,7 +246,7 @@ IfWhen("the '' is found in '' at least '' times", function(ele, arr, times)
 	local found = constructor(0)
 	local one = constructor(1)
 	for _,v in pairs(list) do
-		if v == obj then found = found + one end
+		if type(v) == type(obj) and v == obj then found = found + one end
```
jaromil, line 249: this part coul dleverage `:octet()` on `iszen(type( ) )` positive results, if float cast to octet is possible.

jaromil:
merging for now, spawned issue #453

src/lua/zencode_array.lua
albertolerda: `:octet()` doesn't see the difference in type. We risk to have the octet which represent the string "1" to be equal to the float 1 (we don't want different type that are equal when compared)

Issue #446, opened 2022-08-10
ZKP Verify proof validation


andrea-dintino:
Code: 

```
Scenario credential: anyone verifies the proof
Given that I have a 'issuer public key' inside 'The Authority'
Given I have a 'credential proof'
When I aggregate all the issuer public keys
When I verify the credential proof
Then print the string 'The proof matches the public_key! So you can add zencode after the verify statement, that will execute only if the match occurs.'

```

Data:

```
{
	"The Authority": {
		"issuer_public_key": {
			"alpha": "Ci3RNr3mQgZ+vdhwEokPnq7djLCyoTotPpqfBRQZoX8VdKdGBhkGoIAKfwGQSZH0CG1ZyssgkjW/SRKElf5sHhOjZWzt9xhfwv70RWy8rZVCxQjlYsq+ziFVPPj/DEbLD+1UzGq3qD0awV5oWOmbP76ogkBobBrLboBT5cTLiLTqBdFo356Y1LwOFrafSeKQFxkZxsFi6TCXrzKqzJDOo7zPSXmbBQgh14cz9JNdKNxt+vc94qVJCbpdJufLjQeE",
			"beta": "C1PrsjE2/BamDZIlMgqs+X26/RzgzKkcG4gL2QhEvubGwwE1+jOEcZlvn/Tv3htxCr/8ytK6o6QNit9ry83btHtKkXaTmeg2DbvX9A5vix9K6s9Y5kcbj0BDsEAePcu0EOdxtJlFsjOcCmOV/DqD+9zq1uVxQvcqSa25PGAtxVOrSOTAuCVX8enZ2mpiNtRZGIF5kROS3KZk98wPz4QtarzbQwV+TWPnE36/R0twjk/S14xYhGMYvjLPKBXr8YpP"
		}
	},
	"credential_proof": {
		"kappa": "DgqPHKe6eltcdXHsSUSci+sq9BstH7T1pdasBPzM4A2u5yhQsqmHEayLPX59qpcPA6SvfWkd2rIqC2zus4TwumK38g9Lh9p0jSibAfTk7Xsl1HsRcR379U/4Ucdssxb0CDrquiy3KyMcJN+BoR3hFKIO+5i0wDb5/m0BhK/f9dp0QuHQmHn/AdZmFyvuxHGdC9zHsohg4CKCcWA1AS8lBfLVOHsf5rX/nrq5+ZD+DN7lWv+svqg/BHM7pUiCrwNw",
[...]

Issue #444, opened 2022-08-10
Reflow #2 - verify seal


andrea-dintino:
Code: 

```
# This script can also be run by anyone
# The script is used to verify the 'identity' of the 'seal', 
# so to check if certain information was stored in the 'seal'
Scenario 'reflow' : Verify the identity in the seal

# Here we load the signed seal, along with the data used to generate the identity 
Given I have a 'reflow seal'
Given I have a 'string dictionary' named 'Sale'

# Here we generate the 'identity', with data that presumably was the same
# as the one used when the 'seal' was generated
When I create the reflow identity of 'Sale'
When I rename the 'reflow identity' to 'SaleIdentity'

# Below we're verifying if the 'identity' matches, by doing a simple string comparison
When I verify 'SaleIdentity' is equal to 'identity' in 'reflow seal'

Then print the string 'The reflow identity in the seal is verified'

```

Data:

```
{
	"Sale": {
		"Buyer": "Alice",
		"Seller": "Bob",
		"Good": "Cow",
		"Witness": "Carl",
		"Price": 100,
		"Currency": "EUR",
		"Timestamp": 1422779638,
		"Text": "Bob sells the cow to Alice, cause the cow grew too big and Carl, Bob's roomie, was complaining"
	},
	"reflow_seal": {
		"SM": "AwAbmm8ZljE1IZBM5GqwyiRCRDR0H4SylFXOkL4LbLrBAYIqJ7Sk8GyY0Gi2+Wmg3w==",
		"fingerprints": [
			"AxanCv97hgc/eja1hlic8+WCau6IC+uD0WawpTXPZ4RdC0ACiaiA0xXjjP/0RjUmcw==",
			"AhfnbVh5TtNTQWzOSOupx54RaVBnI7DmxrTkk9Z1qIfIaPKk67t+m2C2R+eQdFw4Qw==",
			"Aheh79g/EKQtlZVD7zKXPbLGvbDWgFRarhx9TcC6Uqw3bUWE2XGyFQyiBgRV8Tme8w=="
		],
		"identity": "AgIgKuNr+bD5UCmRPyASnsGae2eaNR6f2hyxebzdSfGimOMLtFTWORs8k6boJYEwHA==",
[...]

PR #454, opened 2022-08-10
reflow CODECS

albertolerda:


albertolerda:
This solves #444 and #446

jaromil:
well done! I couldn't do better, also wasn't sure if base64 everywhere was best choice. zk proofs everywhere can benefit from this approach

PR #473, opened 2022-09-02
External api hash

jaromil:


jaromil:
awesome!! @phoebus-84 this api will be merged and ready to use for hashing files

PR #477, opened 2022-09-08
Mimalloc

jaromil:
new internal memory allocation library for improved security and speed


jaromil:
needs review to make sure we do not have a huge speed decrease in WASM/JS

PR #489, opened 2022-09-20
Refactor pypi release github action

matteo-cristino:


jaromil:
rust test failure seems non-deterministic, happened also earlier today on master, re-run now

PR #494, opened 2022-09-28
fix: zenroom-integration and crypto-integration for check targets

matteo-cristino:


build/tests.mk
```diff
@@ -134,7 +125,6 @@ check-linux:
 	$(call bats, test/determinism)
 	$(call crypto-integration,${test-exec})
 	$(call zencode-integration,${test-exec})
-	cat /tmp/zenroom-test-summary.txt
```
jaromil, line 137: oh thanks! that was an annoying leftover

PR #501, opened 2022-10-05
added sdist build and deployments for pypi

eckelj:
We are starting to support more and more architectures with planemint and would love to deploy sdist packages so that a smooth deployment with new zenroom packages is garanteed. 

Therefore, I adjusted the pypi deployment workflow to contain sdists.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -242,6 +242,20 @@ jobs:
           pip3 install .[test]
       - run: pytest -s tests
 
+  make_sdist:
+    name: Make SDist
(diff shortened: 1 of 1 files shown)


puria (approved):

puria:
> @puria ups, I saw that the workflows are not executed if an external PR is issued. I will redo the PR directly from your repo.

It's fine, I was in the train so didn't merge it, will do now. And yes we are limiting the workflows usage to stay in the free tier!

eckelj:
all good! thx!

PR #503, opened 2022-10-06
fix js to include teardown messages

jaromil:
correct exit code processing


jaromil:
error message changed, now has "execution aborted with error" to distinguish from other future reasons we may report. will update test accordingly.

PR #512, opened 2022-10-25
fix: export of petition tally

matteo-cristino:


jaromil:
well done and thanks @andrea-dintino for spotting it

PR #517, opened 2022-11-08
Fixed the issues of deployable python sdist packages

eckelj:
Added some fixes in python pkg deployments so that the source packages contain all the data needed to build the package on the deployed hosts.

* added prepare.sh script to prepare the more generic python dist creation
* changed GH to use the bindings/python3 folder and the prepare.sh script
* added conditional bindings cleaning for the Makefile, this will we skiped in case of exection with int he python source package
* adjusted setup.py to be executed within the bindings/python3 folder for bdist and sdist
* added MANIFEST.in to contain the prepared src folder to enable sdist binary building on host platforms


puria (commented):
Fine for me, now uploads everything needed to compile on your local machine ;) @matteo-cristino can you give a twist? and if okay approve the PR

PR #529, opened 2022-11-24
feat: cast integer to float

albertolerda:


src/lua/zencode_when.lua
```diff
@@ -593,6 +593,16 @@ When("create the '' cast of strings in ''", function(conv, source)
...
 end)
 
+When("create the float '' cast of integer in ''", function(dest, source)
+	empty(dest)
+	local src = have(source)
+    if type(src) ~= 'zenroom.big' then
+        src = BIG.new(src)
+    end
+    I.spy(src)
+    ACK[dest] = I.spy(F.new(I.spy(BIG.to_decimal(src))))
```
jaromil, line 603: debuggig leftover `I.spy`

PR #538, opened 2022-12-01
added atomic linker flag to support special HW/OS

eckelj:
added atomic linker flag to build zenroom library for the following CPU/hardware

model name	: ARMv7 Processor rev 3 (v7l)
Hardware	: BCM2711
Model		: Raspberry Pi Compute Module 4 Rev 1.0

and the following OS

Raspbian GNU/Linux 11 (bullseye)


jaromil:
Some targets, presumably also iOS, will not have libatomic. Placing it as global will break them, i.e macos:
```
ld: library not found for -latomic
```
I can accomodate this change for the targets where its required.
Can you confirm what `make` targets are you using to build for the architectures you mention?

jaromil:
It still breaks python builds on OSX. If there is a way to single out the case of building on rPI platforms or OSX and add the flag only for rPI or remove it on OSX, we can merge this. I lack experience customizing setup.py and meson scripts to do this.

PR #539, opened 2022-12-02
feat: zencode foreach

albertolerda:


src/lua/zencode.lua
```diff
@@ -636,7 +637,9 @@ function zencode:run()
 	end
 
 	-- EXEC zencode
-	for _, x in pairs(self.AST) do
+	while ZEN.next_instruction <= #self.AST do
+		x = self.AST[ZEN.next_instruction]
```
jaromil: This is OK. Just a note, perhaps TODO: for optimization, to develop a lua iterator, which would save lookup time
https://www.lua.org/pil/7.1.html

test/zencode/foreach.bats
```diff
@@ -0,0 +1,118 @@
...
+When I create the new array
+Foreach 'x' in sequence from 'zero' to 'limit' with step 'one'
+When I insert 'x' in 'new array'
+EndForeach
+Then print 'new array' as 'string'
+EOF
+    save_output "foreach_big.out"
+    assert_output '{"new_array":["-5","-3","-1","1","3","5","7","9","11"]}'
+}
+
```
jaromil, line 118: would be good to test multiple When statements inside foreach blocks with a bit more variety

jaromil (approved):
awesome work, very well done

PR #541, opened 2022-12-02
refactor(w3c): ♻️  all public key to base58

matteo-cristino:


src/lua/zencode_w3c.lua
```diff
@@ -20,88 +20,55 @@
...
-    if doc.verificationMethod then
-	vm = deepmap(export_vm, doc.verificationMethod)
-	doc.verificationMethod = nil
-    end
     res = deepmap(O.to_string, doc)
-    res.verificationMethod = vm
     return res
 end
 
 local function import_verification_method(doc)
```
puria, line 38: Some inspirational patterns: 
http://lua-users.org/wiki/SwitchStatement
and
https://stackoverflow.com/questions/37447704/what-is-the-alternative-for-switch-statement-in-lua-language

jaromil:
💪🏽

PR #550, opened 2022-12-12
feat(w3c): 🔍️ more checks during did documents import

matteo-cristino:


src/lua/zencode_w3c.lua
```diff
@@ -23,9 +23,34 @@ local ETHEREUM_ADDRESS = "ethereum_address"
...
+    ZEN.assert(doc['@context'], 'Invalid DID document: @context not found')
+    local did_components = strtok(doc.id, '[^:]*')
+    -- schme: did
+    ZEN.assert(did_components[1] == 'did',
+	       'Invalid DID document: invalid scheme')
+    -- method name: a-z/0-9
+    ZEN.assert(not did_components[2]:match('[^%l%d]'),
+	       'Invalid DID document: invalid method-name')
+    -- mathod specific identifier: a-z/A-Z/0-9/"."/"-"/"_"/"%" HEXDIG HEXDIG
+    local error_msg = 'Invalid DID document: invalid method specific identifier'
```
jaromil, line 37: maybe include also base64 and url64 chars for future uses

src/lua/zencode_w3c.lua
```diff
@@ -23,9 +23,34 @@ local ETHEREUM_ADDRESS = "ethereum_address"
 local EDDSA_PUBLIC_KEY = "eddsa_public_key"
 
 local function import_did_document(doc)
-    -- id must be always present in DID-documents
+    -- id and @context must be always present in DID-documents
     ZEN.assert(doc.id, 'Invalid DID document: id not found')
-    -- all the other fields are optional and imported as a string
+    ZEN.assert(doc['@context'], 'Invalid DID document: @context not found')
+    local did_components = strtok(doc.id, '[^:]*')
```
jaromil, line 29: also add a check for size with some plausible defaults, i.e: 
- doc.id will never be longer than 256 bytes
- doc.@context will never be longer than 512 bytes
... any other limit in size that can be applied safely

PR #552, opened 2022-12-12
fix(dictionary): 🐛 move from and copy in dictionaries/arrays statements

### LINKED ISSUE #544: move elements from dictionaries to arrays
When running the following zencode
```gherkin
Given I have a 'string dictionary' named 'dict'
When I create the 'string array' named 'result'
When I move 'test' from 'dict' to 'result'
Then print the 'result'
```

with input
```json
{
	"dict": {
		"test": "test"
	}
}
```

The result should be 
```json
{
   "result": [
      "test"
   ]
}
```
Instead it is
```json
{
   "result": {
      "test": "test"
   }
}
```

matteo-cristino:
Fix #544

### DIFF
--- a/src/lua/zencode_dictionary.lua
+++ b/src/lua/zencode_dictionary.lua
@@ -283,13 +283,14 @@ When("for each dictionary in '' append '' to ''", function(arr, right, left)
 	end
 end)
 
-local function move_in(src_value, src_name, dest)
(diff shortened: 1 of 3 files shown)


jaromil:
Can you please have a look at #555 and extend this PR with the local `function copy_or_move_in()` to support all cases including copy in?
Probably simplier than merging conflicts, then I can delete my PR unmerged

PR #566, opened 2022-12-23
feat(w3c): ✨ extract pubkeys from verification method

### LINKED ISSUE #564: simple w3c scenario statement to extract pubkey from verification method
```
When I create the verificationMethod of 'signer_did_document'
and I pickup a 'ecdh_public_key' from path 'verificationMethod.ecdh_public_key'
and I create the json of 'request did document'
and I verify the 'json' has a ecdh signature in 'ecdh signature' by 'ecdh public key'
and I remove the 'verificationMethod'
```

Could be simplified having a w3c scenarion statement like
```
When I create the 'ecdh public key' from 'signer did document'
```

matteo-cristino:
Fix #564

### DIFF
--- a/src/lua/zencode_w3c.lua
+++ b/src/lua/zencode_w3c.lua
@@ -297,3 +297,24 @@ When(
                                           encoding = 'complex' })
     end
 )
+
+When(
+    "create the '' public key from did document ''",
+    function(algo, did_doc)
+        local doc = have (did_doc)
+        local pk_name = algo..'_public_key'
+        empty (pk_name)
+        ZEN.assert(doc.verificationMethod, 'verificationMethod not found in '..did_doc)
+        local id = doc.id..O.from_string('#'..pk_name)
+        local i = 1
+        repeat
+            if doc.verificationMethod[i].id == id then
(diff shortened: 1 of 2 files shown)


jaromil:
since this is open already.... another even simplier statement:
```
When I verify the did 'doc' is signed by 'signer'
```
takes directly two dids, the signer one and the one signed, checks that the signature is valid, also eventually that the verificationMethod is the same (algo, curve)

src/lua/zencode_w3c.lua
```diff
@@ -297,3 +333,28 @@ When(
                                           encoding = 'complex' })
     end
 )
+
+When(
+    "create the '' from ''",
```
jaromil, line 338: should be less generic:
```
create the public key '' from did document ''
```

matteo-cristino: Yes, you are right, it should be less generic. Two doubts:
- This statement creates an object called `public key` right? 
- Moreover, since it creates only public keys, this statement will not be used to extract  the `ethereum adrress`?

PR #583, opened 2023-01-12
fix: Use sha512 from milagro in ed25519-donna

albertolerda:


jaromil:
well done! just the random left to go

PR #597, opened 2023-02-08
fix(given): 🐛 given nothing do not consider http_request in data if Rule caller restroom-mw is used

matteo-cristino:


albertolerda:
See https://github.com/dyne/restroom-mw/pull/238 for an implementation inside restroom.

PR #600, opened 2023-02-24
fix(cbor): remove CBOR from docs and rules

### LINKED ISSUE #596: Is CBOR format still available?
Running the following script
```gherkin
Rule output format cbor
Scenario 'ecdh': Create the key

Given nothing
When I create the ecdh key
Then print the 'keyring'
```
Then I get the following error:
```
1675765268736E[!] ERROR:
1675765268736E[!] PARSE: /zencode_data.lua:377: attempt to index a nil value (global 'CBOR')
1675765268736E[!] Execution aborted
1675765268736E[!] Execution aborted with errors.
```
Does Zenroom still support input and output CBOR format? If it is not supported anymore then both the [get_format function](https://github.com/dyne/Zenroom/blob/master/src/lua/zencode_data.lua#L372-L382) and the [documentation](https://dev.zenroom.org/#/pages/zencode-rules) should be updated.

jaromil:
fixes #596

### DIFF
--- a/docs/pages/zencode-cookbook-given.md
+++ b/docs/pages/zencode-cookbook-given.md
@@ -13,7 +13,7 @@ Link file with relative path:
 # The *Given* phase: reading and verifying input
 
 The *Given* keyword marks the first phase of Zencode execution, where input is read and processed and first internal variables are set. More precisely: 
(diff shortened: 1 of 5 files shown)


matteo-cristino (commented):
Seems perfect to me! ✨

PR #612, opened 2023-03-13
feat: use strtok from c (fix #608)

### LINKED ISSUE #608: Secure strtok in C
Substitute the `strtok` used in Lua with an implementation in C that leverages the LIBC's strtok and creates a lua table.

The strtok function is involved in some security checks and should be solid rather than based on a Lua's regexp.

albertolerda:


jaromil:
Well done!!

just beware

image

There are a few more places where `strtok` is used, sometimes without a second argument (default is space)

albertolerda:
Are you in the right branch? To me, it looks like they are fixed..

PR #615, opened 2023-03-15
feat(hash): ✨ zencode statement to create an hash to point

matteo-cristino:
Also deprecated the hash to point of each object in an array with the following message:
```
[W]  DEPRECATED:
create the hash to point '' of each object in ''
use instead
Foreach 'element' in 'array'
When I create the hash to point of 'array'
and I move 'hash to point' in 'array of hashes'
EndForeach
```

### DIFF
--- a/src/lua/zencode_hash.lua
+++ b/src/lua/zencode_hash.lua
@@ -51,20 +51,36 @@ When(
     end
 )
 
-When(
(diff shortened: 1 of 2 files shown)


jaromil:
well done also the deprecation!!!

PR #620, opened 2023-03-29
fix(ethereum): 🐛 accept any zenroom type in the store of data

### LINKED ISSUE #619: Broadcast ECP (and ECP2) points on ethereum
I'm trying to store on an ethereum blockchain an hash to point of some data, but when using the following piece of code
```gherkin
When I create the hash to point 'ecp' of 'broadcast_data'
When I create the ethereum transaction to 'storage contract'
When I use the ethereum transaction to store 'hash to point'
```
The followinf error is thrown:
```
[!] /crypto_ethereum.lua:288: attempt to get length of a zenroom.ecp value (local 'src')
```
Where the relative line of code is:
https://github.com/dyne/Zenroom/blob/f195a10244ef4afeb7d673f6e172811da1ab4450/src/lua/crypto_ethereum.lua#L277-L288

Thus i think that an implementation of `__len` function for ECP and ECP2 is needed. Do they have a fixed length and are padded in the case they are shorter? In this case this function can simply be the return of such number.

matteo-cristino:
Fix #619

### DIFF
--- a/src/lua/crypto_ethereum.lua
+++ b/src/lua/crypto_ethereum.lua
@@ -285,6 +285,9 @@ function ETH.make_storage_data(src)
    offset = O.from_hex('0000000000000000000000000000000000000000000000000000000000000020')
 
    -- length as a 256 unsigned integer
+   if iszen(type(src)) then
+      src = src:octet()
+   end
    bytLen = INT.new(#src):octet()
    paddingLength = 32-#bytLen
    paddingLen = O.zero(paddingLength)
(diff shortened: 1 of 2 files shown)


src/lua/crypto_ethereum.lua
```diff
@@ -285,6 +284,9 @@ function ETH.make_storage_data(src)
    offset = O.from_hex('0000000000000000000000000000000000000000000000000000000000000020')
 
    -- length as a 256 unsigned integer
+   if iszen(type(src)) then
```
jaromil, line 287: I am not 100% sure of the way arguments are used in lua, as references or not. However it is safer here to make src as a const and instead of overwriting its contents to create a new local variable like src_o

jaromil (changes requested):

PR #621, opened 2023-03-30
fix(ethereum): 🐛 import of chainid

matteo-cristino:


jaromil:
much better, I remember looking with suspicion at those lines of code

Issue #607, opened 2023-03-10
Easy verify length of array is less or equal


jaromil:
Simplify check in many statements in which we need to check that an array is less or equal
```
and I create the length of 'array'
and number 'length' is less or equal than '2'
```

This is used in secure checks for instance splitting over `_` underscore to see if a did:domain has an `_A` for admin flag

jaromil:
```
When I verify the length of '' is less or equal "
When I verify the size of '' is less or equal "
```

length and size equivalence?

matteo-cristino:
Done in #624

Issue #606, opened 2023-03-10
Easy copy and rename an element inside an array or dictionary


jaromil:
Faster extraction by copying an element inside a table using the `from` particle in Zencode

Example:
```
 When I create the 'signer_sda' from '3' in 'array'
```
array's keys are always numbers, or for dictionary (autodetected)
```
When I create the 'signer_sda' from 'signer' in 'addressbook'
```
autodetection made based on luatype of key and CODEC.zentype

matteo-cristino:
Done in #624

PR #624, opened 2023-03-31
feat(array): size comparison in one zencode statement

jaromil:
review statements for array size and add direct comparison for one-liner

makes much more readable code i.e. in DID implementation


jaromil:
Should address two issues:
- https://github.com/dyne/Zenroom/issues/606
- https://github.com/dyne/Zenroom/issues/607

jaromil:
Looks perfect! again many thanks for finishing what I've barely started <3

Issue #633, opened 2023-05-03
Test for signature fails for the wrong reason


albertolerda:
The test `Fail verification on a different message` should fail because we pass a message and the signature of another message, but if we use `assert_output` we can see that it is actually failing because he cannot read a file
```
reading DATA from file: sign_pubkey.json
Error opening No such file or directory
```

Test run with `test/bats/bin/bats test/zencode/generic_schnorr.bats`

RebeccaSelvaggini:
The same error occurs in the test `Fail verification on a different public key`. 
By using `assert_output ''` the test fails because it cannot read a file:
``` 
reading DATA from file: wrong_pubkey.json
Error opening No such file or directory
 ```
Test run with `test/bats/bin/bats test/zencode/generic_schnorr.bats`

**Comment by matteo-cristino** [comment]
I think that the bats `run` command reads inputs from `$BATS_FILE_TMPDIR` while the `save_output` and `save_asset` functions store the files in `$BATS_SUITE_TMPDIR`, indeed changing the latter functions to save in `$BATS_FILE_TMPDIR` the test fails with the right error (it can be seen also using `assert_success` instead of `assert_failure`). There was a reason why this two functions stores the files there?

**Comment by jaromil** [comment]
@matteo-cristino according to bats docs, file and suite have different scope and will be deleted at different stages. I believe that a file may contain multiple suites. There may be a bug in the way our tests were engineered, I got confused quite often.

**Comment by albertolerda** [comment]
`assert_failure` should also accept a string and verify that the failure was due to the expected problem.

**Comment by matteo-cristino** [comment]
From the bats documentation:
* `$BATS_SUITE_TMPDIR` is a temporary directory common to all tests of a suite. Could be used to create files required by multiple tests.
* `$BATS_FILE_TMPDIR` is a temporary directory common to all tests of a test file. Could be used to create files required by multiple tests in the same test file.

I think (to be checked) that we do not access the same file from different test files, thus `$BATS_FILE_TMPDIR` should be fine.

**Comment by matteo-cristino** [comment]
> `assert_failure` should also accept a string and verify that the failure was due to the expected problem.

A solution for that is to use `assert_line` instead of `assert_failure`: this function helps to verify that a command or function produces the correct output. It checks that the expected line appears in the output (default) or in a specific line of it.

For example in the signature failing test:
```
run ...
assert_line '[W]  The signature by Alice is not authentic'
```
[...]

PR #637, opened 2023-05-09
test(zencode): 🧪 refactor for better checks on failing tests

matteo-cristino:


src/lua/crypto_schnorr_signature.lua
```diff
@@ -152,18 +155,18 @@ function schnorr.verify(pk, m, sig)
    if iszen(type(pk)) then pk = pk:octet() end
    --the follwing "lifts" pk to an ECP with x = pk and y is even
    local P = ECP.new(BIG.new(pk))    
-   assert(P, "lifting failed")
+   if not _schnorr_assert(P, "lifting failed") then return false end
```
jaromil, line 158: how about `return _schnorr_assert(...)`

matteo-cristino: you want to exit only in the case _schnorr_assert is false, in the case it is true yo want to continue with all the checks

matteo-cristino:
fix #633

PR #650, opened 2023-05-23
build: fix meson build for python source distribution

matteo-cristino:
In python source distribution the `.git` folder is not present to reduce the package size. A file, named `git_utils`, contains the tag, commit hash and time of the last commit. The `make meson` command now if  `.git` is not found search the needed information inside  `git_utils`.


jaromil:
cheers!!!

PR #653, opened 2023-05-25
feat: ethereum sign and verify from address

Luckydd99:
***NOTE:*** the function creating the signature in this branch uses RANDOM ecdsa.  
It should be substituted with the deterministic one inside the "ecdsa-deterministic" branch.


jaromil:
Amazing, we completed this missing piece and can now sign eth tx all in zenroom! cheers!

I'll review in detail ASAP

src/lua/crypto_ethereum.lua
```diff
@@ -348,7 +373,7 @@ end
 
 local encode_tuple
 
-local function encode(t, arg)
+function encode(t, arg)
```
albertolerda, line 376: This should be kept local, we export in with `ETH.abi_encode = encode`

Luckydd99: At line 230, the function encode is indeed initialised as "local", and then it is declared and defined at 376.

Issue #652, opened 2023-05-25
Import dictionary of schema / ZKP ABC improvement


albertolerda:
The zencode
```
Scenario 'credential': participant aggregates credential signature(s)

Given I have a 'credential_signature_dictionary' named 'cred_sig_dict'

Then print 'cred_sig_dict'
```
With data
```
{
	"cred_sig_dict": {
		"credential_signature": {
			"a_tilde": "Aw2/mA+dfd1SdvC/3WLAsXK5l/XuwY0xN/sLxpeXY+2hI9PMZR1RkDw7nCG6W1xGHw==",
			"b_tilde": "AwgOIhjVfNQc5uQDbsWXHuppPM2QGhkFl/byLZI8JkeGQXf1pn2PH+n7x0TeHYHUPQ==",
			"h": "AhPY7CWI4FraIkfdVSFryiqmmhsrMAUQZ2cU9gx8NagElM6EVmvwkquyGvvmNVfPvw=="
		},
		"AgeAuthority_cred": {
			"a_tilde": "AwosCPFxV+ZaplVejSOP44us+dBnVnxP0pw9paSG6q0dYw4bA5iV97mb+19fjjzqCw==",
			"b_tilde": "AhDAL96HhsUflFMHPTOx1167dss4tauvqZiXqa/MS6jm8sAW2thUZV8g5hGcd4BCeA==",
			"h": "AwvH5BaupA168IC18UOveDgbXHrEnbwbM8nrcOJB4ffkdcYXEnogltmhq0Na6kz4HQ=="
		}
	}
}
```

Throws the error
```
[!] [string "zencode_credential"]:127: Key not found in object conversion: h
```

jaromil:
> Given I have a 'credential_signature_dictionary' named 'cred_sig_dict'
> Throws the error
> [!] [string "zencode_credential"]:127: Key not found in object conversion: h
> ```

This happens because zencode does not support (yet?) to have 'schema + dictionary/array' declared in given. The dictionary/array suffix (last word in string) works only on base encodings (string, base64, etc.) and not on complex schemas.

It is perhaps the time to implement this.

jaromil:
fixed

andrea-dintino:
Verification missing: 

Zencode: 

```
Scenario credential: anyone verifies the proof
Given I have a 'issuer_public_key_dictionary' named 'issuer_public_key_dict'
Given I have a 'credential proof' inside 'Participant'
[...]

PR #655, opened 2023-05-25
fix(schema): refactoring complex schema code

jaromil:
all code handling schemas that have unique functions for import and export needs a bit of a cleanup and small refactoring, this is also removing the 'keyring' as special case and including it as complex schema.

address #652


jaromil:
All tests pass except `zencode_float`, where a JSON number with a floating point is imported from a given integer statement without error. I need help to analyse if this is not desired and how to put safety in there @albertolerda

PR #656, opened 2023-05-30
Ethereum address checksum encoding

RebeccaSelvaggini:


jaromil:
Well done!!! Can you also clearly state in the documentation that for checking the validity of an imported ethereum address one needs to read it also as string?

```
Given I have a 'string' named 'ethereum address' 
When I verify the ethereum address 'ethereum address' is valid
```

This is perhaps a good start for an example (taken from the test unit). The documentation should show a real-case scenario in which an address is imported, checked and then used to create a transaction. Reading the code now I'm not sure if its the case that the address needs to be imported twice, as schema and as string...

jaromil:
last force push is just the conflict resolution for minor conflicts caused by recent changes merged in master.

jaromil:
well done!!!

PR #657, opened 2023-05-30
fix: create pk doesn't change keyring (fix #634)

### LINKED ISSUE #634: `When I create the eddsa public key with secret key ''` should not change the keyring
If I run the script
```
Scenario 'eddsa': boh
Given I have the 'keyring'
Given I have the 'base58' named 'another_key'
When I create the eddsa public key with secret key 'another_key'
Then print the 'keyring'
```
with keys
```
{
	"keyring": {
		"bitcoin": "L2rYniNJRzGu3V5SJciVYeEY8eGBdpJMrwM5v3oTo6efvZA3i3XM",
		"ecdh": "4Aiyv8ZlUlV/VIbkXJw/isoMeIAyrCEjDxJGB50avtI=",
		"eddsa": "6QTJ78cqECjgPS1MVESXjyqxcgCwqTUatkqKYDDxB1CG",
		"ethereum": "2797ea055d708e4287d3cceb59fe11bae8fe6cdd54852fdb11fe6fd3a34d5530",
		"reflow": "CBh8Ovfq+XIPTL16IhNS3Bq29putPzTfsgHcN1cRfoU=",
		"schnorr": "OMeE4MZGKIIbvSuwqnwYVF+4+jqGtBD2gwsX57TZ8sY="
	}
}
```
and data
```
{
	"another_key": "BfUdMQn6DnZzFS2tFUK7TxS8VBeujCDtmqEfFGdTd6FB"
}
```

It fails because it tries to replace the key in the keyring
```
[W]  Cannot overwrite existing key: eddsa
[W]  [!] [string "zencode_eddsa"]:66: Cannot overwrite existing key: eddsa
[!]  ERROR:
[!]  EXEC: [string "zencode"]:774: When I create the eddsa public key with secret key 'another_key'
```

albertolerda:


matteo-cristino:
The same statement is also used in the [qp scenario](https://github.com/dyne/Zenroom/blob/master/src/lua/zencode_qp.lua) for dilitihum, kyber and ntru prime and in the [schnorr scanrio](https://github.com/dyne/Zenroom/blob/master/src/lua/zencode_schnorr.lua#L83-L90). Maybe it's worth fixing them all

PR #662, opened 2023-06-01
fix: 🐛 add schematype field to describe which schemas can be modified

matteo-cristino:


jaromil:
OK this looks good, but how do we know that the schema export handles additional object insertion? if left implicit, one cannot be sure it does or not, which depends from the way its implemented. In case of DIDs, the implementation makes it possible (we may call it and "`open schema`") i.e:
```lua
local function export_did_document(doc)
   local t = luatype(doc)
   return deepmap(O.to_string, doc)
end
```
as well `import_verification_method` and `import_did_document` contemplate this possibility.

Other implementations may not support it, for instance the `import_credential_request_f`:
```lua
local function import_credential_request_f(obj)
   local req = {
      sign = {
     a = ZEN.get(obj.sign, 'a', ECP.new),
     b = ZEN.get(obj.sign, 'b', ECP.new)
      },
      pi_s = {
          rr = ZEN.get(obj.pi_s, 'rr', INT.new, O.from_base64),
          rm = ZEN.get(obj.pi_s, 'rm', INT.new, O.from_base64),
          rk = ZEN.get(obj.pi_s, 'rk', INT.new, O.from_base64),
          commit = ZEN.get(obj.pi_s, 'commit', INT.new, O.from_base64),
      },
      commit = ZEN.get(obj, 'commit', ECP.new),
      public = ZEN.get(obj, 'public', ECP.new)
   }
   ZEN.assert(
      CRED.verify_pi_s(req),
      'Error in credential request: proof is invalid (verify_pi_s)'
   )
   return req
end
```

therefore we need to add a `schematype` (or `st`) field in the `CODEC` which says `open` (or `o`) in case this is possible.

jaromil:
Actually I managed to implement a less invasive change, just adding a field to the schema (not `CODEC`) resolved at runtime (property not exported in codec), so far only implemented for did_document:
```lua
ZEN.add_schema(                                                                     |      pickin(struct, name, sc)
    {                                                                               |      assert(TMP.name,  'No name found in: '..name)
        did_document = { import = import_did_document,                              |      Iam(O.to_string(operate_conversion(TMP)))
                         export = export_did_document,                              |      ZEN.CODEC[name] = nil -- just name string
                         schematype = 'open' },
```

any other schema that is not marked `open` and will have an object moved insides will return error.

src/lua/zencode_data.lua
```diff
@@ -387,6 +387,14 @@
     return nil
  end
 
+
+ -- CODEC format:
+ -- { name: string,
+ --   encoding: encoding of object data, both basic and schema
+ --   zentype:  zencode type: 'e'lement, 'a'rray or 'd'ictionary
+ --   schema: schema name used to import or nil when basic object
+ --   st: schema type properties, so far just 'o'pen or none
```
jaromil, line 396: this slipped through, but is not needed anymore

PR #665, opened 2023-06-05
ci: 👷 align releases and remove tag in case of failure

matteo-cristino:


puria (approved):

matteo-cristino:
In this PR I have aligned PyPI, npm and github releases. In this way if one build fail no release will be made in any of them. Moreover, to keep their version aligned with the github version, in case of failure of one of the builds no release will be made and the tag released on github will be eliminated. This works also with jobs cancellation so that if a new PR is merged during the build phase the CI will be stopped and the github tag will be eliminated. This should help to keep all the version aligned.

Also the js release is build after the semantic versioning so the tag now match with the commit hash.

puria (approved):

andrea-dintino:
@jaromil can we pretty please merge?

Issue #669, opened 2023-06-09
Create ethereum signature hash


andrea-dintino:
We found that the standard for ethereum signatures (at least based on [etherscan](https://etherscan.io/verifiedSignatures#) + Metamas ) is that signatures are formatted as "signature hash". A signature hash is basically a cat of **0x + r + s + hex(v)**. It would be nice to have a Zencode statement that does it, e.g.:

```
Scenario ethereum: keygen and wallet

Given I am 'participant'
Given I have my 'keyring'
Given I have a 'string' named 'myString'

When I create the ethereum signature hash of 'myString'

then print the 'ethereum signature hash'
```

Input:

```
{
	"myString": "I love the Beatles, all but 3",
	"participant": {
		"keyring": {
			"ethereum": "6b4f32fc48ff19f0c184f1b7c593fbe26633421798191931c210a3a9bb46ae22"
		}
	}
}
```

output

```
{
"'ethereum signature hash": "0xed8f36c71989f8660e8f5d4adbfd8f1c0288cca90d3a5330b7bf735d71ab52fe7ba0a7827dc4ba707431f1c10babd389f658f8e208b89390a9be3c097579a2ff1B" 
}
```

PR #671, opened 2023-06-12
feat: zencode ethereum signature hash

RebeccaSelvaggini:
See [#669](https://github.com/dyne/Zenroom/issues/669).

In src/lua/zencode_ethereum.lua now we have:
- [export function](https://github.com/dyne/Zenroom/blob/eth-sig/src/lua/zencode_ethereum.lua#LL110C1-L115C4)  of ethereum signatures, now the standard is to print the signature as a single hexadecimal string;
- [statement](https://github.com/dyne/Zenroom/blob/eth-sig/src/lua/zencode_ethereum.lua#LL364C1-L371C2) for the creation of the ethereum signature, now we have just one statement for the signature, then it can be printed both as a string, or as an array {r,s,v} using the statement '`Then print the 'ethereum signature' as 'hex'`;
- [statement](https://github.com/dyne/Zenroom/blob/eth-sig/src/lua/zencode_ethereum.lua#LL373C1-L385C5) for the verification that takes as input both the string and the array encoding of the signature.


jaromil:
Hi! In this PR There is a lot of re-indenting which should not be there (makes it hard to track changes). Indentation is a different thing than refactoring and should be made in its own commit. In this case is not desirable at all.

PR #680, opened 2023-06-20
feat: call generic smart contract from zencode

albertolerda:


src/lua/zencode_ethereum.lua
```diff
@@ -388,3 +388,21 @@ IfWhen("verify the '' has a ethereum signature in '' by ''", function(doc, sig,
...
     )
 end)
+
+When("use the ethereum transaction to run '' using ''", function(m, p)
+
+    local method = have(m)
+    local params = have(p)
+    ZEN.assert(
+        method.name and method.input and type(method.input) == "table",
+       'method must contain the keys method and input'
```
jaromil, line 398: maybe meant name and input? best also indicate: name (string) and input (table)

src/lua/zencode_ethereum.lua
```diff
@@ -388,3 +388,21 @@ IfWhen("verify the '' has a ethereum signature in '' by ''", function(doc, sig,
...
+    )
+    local input = deepmap(function(o)
```
jaromil, line 400: here is clear that method.input is supposed to be an array (or dictionary?) of strings, so we should also check the CODEC records for that

jaromil (changes requested):

src/lua/zencode_ethereum.lua
albertolerda: `input` is a key inside the method dictionary, so it doesn't have a CODEC.
Should we introduce a schema for a method? This is not something standard, it is a dictionary introduced by us
```
{
   name: string,
   input: [string],
   output: [string]
}
```

jaromil (approved):

PR #690, opened 2023-06-29
fix: pick random in array with one element

albertolerda:


jaromil:
aarg, off-by-one, thanks for checking this out

Issue #681, opened 2023-06-21
Keyword "use" in scenario ethereum


andrea-dintino:
# Situation

We have began to use the keyword "use" in the ethereum scenario: it is used when a previously created transaction is modfied to address a certain purpose, namely: 
- send eth/ERC20
- transfer ER721
- call a specific solidity SC (to store data, we used it for EBSI)
- call a generic solidity SC 

All the Zencode statements are found here:
https://github.com/dyne/Zenroom/blob/master/test/zencode/ethereum.bats
https://github.com/dyne/Zenroom/blob/master/src/lua/zencode_ethereum.lua

# Considerations
- the keyword "use" appears only in zencode_ethereum.lua
- the output of each statement containing "use" is a modified version of the previously created transaction
- the Zenocde statements that called specific solidity SC (e.g. the storage one), can be deprecated and the Zencode to call the generic solidity SC can be used instead 

# Possible solutions
- Keep the "use" keyword, possibly deriving syntax rules a posteriori 
- Deprecate the Zencode containing "use" and replace them with Zencode containing "create", which would require some work to align the data used internally by the Zencode statements

jaromil:
Thanks for opening this issue! Let's discuss this in a meeting seeking possibilities to align the ethereum scenario as much as possible to a generic semantic. The `use` verb seems to always be auxiliary to `store` and there is perhaps an opportunity for simplification.

On a side note also a statement like:
```
When I create the string from the ethereum bytes named 'data'
```
seems it should be deprecated in favor of using an `ethereum bytes` schema.

jaromil:
We are de-facto keeping the `use` verb, present in a number of recent cases already. Closing for now.

PR #697, opened 2023-06-30
feat: create the eth abi decoding

albertolerda:


jaromil:
Related to issue https://github.com/dyne/Zenroom/issues/681 we need to see if `used` is appropriate, perhaps better use `stored`?

albertolerda:
It is actually quite different, the keyword "using", in this case, is used like in the statements `compose the secret using ''`, `create the ethereum abi encoding of '' using ''`, `create the jws signature using the ecdh signature in ''`, .... (it is not a verb, just specify a variable)

In the other PR, the keyword `use` is specific to ethereum and tells how to fill the field data/value of the `ethereum transaction` object

andrea-dintino:
@jaromil we use the keyword 'using' in 7 statements already (include scenario 'secshare'):

https://gyazo.com/a481c21abcae97030f88a8b5391e7c6f

jaromil:
ACK! thanks

PR #700, opened 2023-07-07
feat: create and verify jwt hs256

albertolerda:


test/zencode/w3c.bats
```diff
@@ -701,3 +701,39 @@ EOF
...
+    },
+	"password": "password"
+}
+EOF
+    cat <<EOF | zexe jwt_hs256.zen jwt_hs256.data
+Scenario 'w3c': did document manipulation
+Given I have a 'string dictionary' named 'payload'
+Given I have a 'string' named 'password'
+
+When I create the jwt hs256 of 'payload' using 'password'
```
jaromil, line 720: Are we sure we want to call this `jwt hs256`?

albertolerda: hs256 is the cryptographic schema we are using, I could call it only `jwt` and if, in future, we implement other schemas we have to rename it before using another

PR #704, opened 2023-07-17
feat(pvss): crypto_pvss.lua with zencode and tests

RebeccaSelvaggini:


src/lua/zencode_pvss.lua
```diff
@@ -0,0 +1,227 @@
...
+local function export_verified_shares_f(obj)
+    local res = {["valid_indexes"] = {}, ["valid_shares"] = {}}
+    for i = 1, #obj.valid_indexes do
+        res.valid_shares[i] = obj.valid_shares[i]:zcash_export()
+        res.valid_indexes[i] = obj.valid_indexes[i]:decimal()
+    end
+    return res
+end
+
+local function check_inputs(pks, commitments, encrypted_shares, proof, t, n)
```
jaromil, line 100: No need for a separate local function, since this is used only once in the scope of this file, the code can be included directly in the verify statement.

src/lua/zencode_pvss.lua
```diff
@@ -0,0 +1,227 @@
...
+    ZEN.assert(t <= n, "Quorum is bigger than total")
+
+    ZEN.assert(type(pks) == 'table', "'pvss participant pks' is not a table")
+    ZEN.assert(type(commitments) == 'table', "'pvss commitments' is not a table")
+    ZEN.assert(type(encrypted_shares) == 'table', "'pvss encrypted shares' is not a table")
+    ZEN.assert(type(proof) == 'table', "'pvss proof' is not a table")
+
+    ZEN.assert(#pks == n, "'pvss participant pks' is of wrong length")
```
jaromil, line 108: These checks seem fairly important for input validation, therefore the overloaded operator `#` should not be used to check the length of tables. Use of `pairs` in a iteration is more trustworthy, see https://stackoverflow.com/questions/2705793/how-to-get-number-of-entries-in-a-lua-table

For example:
```
function tablelength(T)
  local count = 0
  for _ in pairs(T) do count = count + 1 end
  return count
end
```

src/lua/zencode_pvss.lua
```diff
@@ -0,0 +1,227 @@
...
+
+end)
+
+-- Anyone verifies the encrypted shares created by the issuer
+When("verify the pvss public shares with '' quorum ''", function(num,thr)
+    local pvss_public_shares = have'pvss public shares'
+    local t = tonumber(have(thr):decimal())
+    local n = tonumber(have(num):decimal())
+
+    check_inputs(pvss_public_shares.public_keys, pvss_public_shares.commitments, pvss_public_shares.encrypted_shares, pvss_public_shares.proof, t, n)
```
jaromil, line 179: replace this with code in check_inputs

src/lua/zencode_pvss.lua
```diff
@@ -0,0 +1,227 @@
...
+    empty'pvss secret share'
+    ACK.pvss_secret_share = output
+    new_codec('pvss secret share')
+end)
+
+-- Each participant verifies the shares of the others
+When("create the pvss verified shares from ''", function(list)
+    local dec_shares = have(list)
+    local valid_shares, valid_indexes = PVSS.verify_decrypted_shares(GENERATORS, dec_shares)
+    for i = 1, #valid_indexes do
```
jaromil, line 207: use of a `pairs` iteration here is faster and more secure

src/zen_ecp.c
```diff
@@ -779,6 +806,8 @@ int luaopen_ecp(lua_State *L) {
 		{"mul", ecp_mul},
 		{"validate", ecp_validate},
 		{"prime", ecp_prime},
+		{"rhs", ecp_rhs},
+		{"zcash_export", ecp_zcash_export},
```
jaromil, line 810: this can be simply a `:zcash()` method of ECP (since it needs always ECP as 1st arg).

If there are both import and export functions then we use `:from_zcash()` and `:to_zcash()` as method names in zenroom.

src/zen_octet.c
```diff
@@ -1976,6 +1976,7 @@ int luaopen_octet(lua_State *L) {
 		{"to_array",  to_array},
 		{"to_octet",  to_octet},
 		{"to_bin",    to_bin},
+		{"zcash_topoint", zcash_topoint},
```
jaromil, line 1979: This should be implemented as `ECP.from_zcash(octet)` and `ECP2.from_zcash(octet)`

I am aware this doesn't only concerns this PR since zcash point encoding was introduced in bbs, but perhaps now is a good time to normalize this interface. It will make also much more readable code where :zcash_topoint() is being used, making it explicit that an ECP is returned, when the method is `ECP.from_zcash(input_octet)`

jaromil (changes requested):

jaromil (approved):

jaromil:
Excellent! many thanks. Lets meet up ASAP to make some meaningful squash of related commits and merge this

src/zen_octet.c
```diff
@@ -2017,7 +2018,7 @@ int luaopen_octet(lua_State *L) {
 		{"charcount", charcount},
 		{"rmchar", remove_char},
 		{"compact_ascii", compact_ascii},
-		{"zcash_topoint", zcash_topoint},
+		// {"zcash_topoint", zcash_topoint},
```
albertolerda, line 2021: Why did we move zcash from zen_octet to zen_ecp/zen_ecp2? 
When I have an octet, I may not be aware of whether it is an ECP or an ECP2, and the algorithm is the same for both. Before I could import the zcash octet and then check its type (it could be both `zenroom.ecp` and `zenroom.ecp2`).
Moreover, it increases code duplication, because we have almost the same function both in zen_ecp and zen_ecp2.

albertolerda: If we want also `ECP.from_zcash` that also asserts the type of the decoded object, I would remove the static from zcash_topoint, make it extern in the other file and add a function that calls it and assert the type of the decoded object immediately after.

jaromil: Resons for this:
1. readability: by making explicit returned type, the lua code is much more readable in absence of type checks
2. awareness: who writes lua code is forced to declare ECP type
3. compliance: en/coding functions are all from_/to_ form in zenroom's lua

code duplication may be fixed, but I am not in favor of making conversion types implicit and possibly concealed, as this may lead to mistakes that are very hard to detect in implementations.

PR #705, opened 2023-07-20
fix(dictionary): 🐛 move element in a table of schemas

matteo-cristino:


jaromil:
Subtle important fix, thanks!

PR #725, opened 2023-08-28
feat: python binding uses `zencode-exec`

albertolerda:


.github/workflows/main.yml
```diff
@@ -543,6 +411,47 @@ jobs:
             src/libzenroom_dll.lib
             src/zenroom.dll
 
+  changes:
```
matteo-cristino, line 414: About this job:
* id should be more meaningfull, like `python-changes`
* `name` is missing, helpfull for a better understading looking at the CI result
* this action should be performed only for commits, add dependency to semantic release with release equal to True and remove `permissions`
* There is only one output, but it is not present in `outputs`, remove backend/fronted and add instead `python3`
* Remove comments

.github/workflows/main.yml
```diff
@@ -543,6 +411,47 @@ jobs:
...
+    - uses: actions/checkout@v3
+    # For pull requests it's not necessary to checkout the code
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        filters: |
+          python3:
+            - 'bindings/python3/**'
+
+  python-sdist:
```
matteo-cristino, line 433: After the fix to the previous job this can depend only on it

.github/workflows/main.yml
```diff
@@ -565,7 +474,7 @@ jobs:
   npm-release:
     name: 📦 npm release
     runs-on: ubuntu-latest
-    needs: [js-build-release, python-build-check, python-build-check-armv7, python-sdist, build-release-on-osx, build-release-on-ubuntu]
+    needs: [js-build-release, python-sdist, build-release-on-osx, build-release-on-ubuntu]
```
matteo-cristino, line 477: should not depends on python-sdist since it does not run always, but I have to think about this since it the logic is not so obvious in this part

.github/workflows/main.yml
```diff
@@ -602,7 +511,7 @@ jobs:
 
   draft-binary-release:
     name: 📦 Pack release
-    needs: [semantic-release, js-build-release, python-build-check, python-build-check-armv7, python-sdist, build-release-on-osx, build-release-on-ubuntu]
+    needs: [semantic-release, js-build-release, python-sdist, build-release-on-osx, build-release-on-ubuntu]
```
matteo-cristino, line 514: same as above

.github/workflows/main.yml
```diff
@@ -627,7 +536,7 @@ jobs:
 
   remove-tag-on-fail:
     name: 🧻 Remove tag on fail
-    needs: [semantic-release, js-build-release, python-build-check, python-build-check-armv7, python-sdist, build-release-on-osx, build-release-on-ubuntu]
+    needs: [semantic-release, js-build-release, python-sdist, build-release-on-osx, build-release-on-ubuntu]
```
matteo-cristino, line 539: same as above

puria (approved):
🦑

PR #731, opened 2023-08-31
feat: rust spawns zencode-exec

albertolerda:


bindings/rust/src/lib.rs
```diff
@@ -38,95 +30,72 @@ impl fmt::Display for ZenError {
...
-    script: impl AsRef<str>,
-    conf: impl AsRef<str>,
-    keys: impl AsRef<str>,
-    data: impl AsRef<str>,
-) -> Result<ZenResult, ZenError> {
-    exec_f(c::zencode_exec_tobuf, script, conf, keys, data)
+    zen_input.push_str(&general_purpose::STANDARD.encode(&extra));
+    zen_input.push_str("\n");
+
+    zen_input.push_str(&general_purpose::STANDARD.encode(&context));
```
jaromil, line 58: shouldn't we define base64 as the encoding here, rather than relying on a standard which may as well change?

PR #739, opened 2023-10-02
feat: print line number of ignored statements

jaromil:
This adds the line number of zencode that aren't recognized and eventually ignored by the parser.

Line number excluded rules and scenario directives.

Tests aren't yet fixed, pending review.


jaromil:
@jaromil @AlbertoLerda as you wished, pending your review and update of tests

PR #740, opened 2023-10-02
Feat/gc generational

jaromil:
Adopt the `generational` garbage collection mode


jaromil:
small glitch here, this includes #738 and should be rebased and merged after that.

PR #741, opened 2023-10-03
test(zencode): 🧪 avoid deprecated statements in tests

matteo-cristino:
* Remove depreacted statements in tests
* Tests now fail in case deprecated statement are found
* Update also documentation


jaromil:
awesome!! very much needed

PR #748, opened 2023-10-18
Breaking changes leading to v4

jaromil:


jaromil:
Thanks for the help aligning all the tests!! I will squash the history to something meaningful and more concise then tag this as BREAKING CHANGE, hence we'll step into 4.0!

PR #749, opened 2023-10-18
fix: 🐛 free allocated octets in case of errors in new_ecp and new_big

matteo-cristino:
Also make bats tests fails in case of memory left allocated


jaromil:
amazing! good catch!!! 🔔

PR #751, opened 2023-10-24
feat(zencode): ✨ rule caller slangroom implementation

jaromil:
Force external zencode statements to the edges (top given and bottom them) enforcing error control when `rule unknown ignore`
`


matteo-cristino:
Implemented in #748 with `Rule unknown ignore`

PR #752, opened 2023-10-24
feat(table): ✨ first implementation of pointer system in zencode

matteo-cristino:
More testing is needed before merging

### DIFF
--- a/build/meson.build
+++ b/build/meson.build
@@ -69,7 +69,7 @@ tests = [
 'generic_es256', 'generic_mldsa44', 'generic_schnorr', 'given',
 'hash', 'http', 'keys', 'kyber', 'mlkem512', 'ntrup', 'numbers',
 'output', 'pack', 'parser', 'planetmint', 'pvss', 'random',
-'reflow','rules', 'scope', 'sd_jwt', 'secshare', 'then', 'time',
+'reflow','rules', 'scope', 'sd_jwt', 'secshare', 'table', 'then', 'time',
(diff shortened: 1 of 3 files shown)


jaromil:
It is safer to use a pointer that is not inside the HEAP, but in the stack (`ZEN.pointer`).

Also we can use the keyword "`pointer`" for all variable names and make it a convention inside Zencode, so that all internal uses of `have()` will return the pointer contents, in case a `ZEN.pointer` is non-nil. This will automatically implement all existing statements  for usage with pointers. Of course this is theoretical, we need some caution and more test coverage.

matteo-cristino:
I have a doubt. In the case `pointer` is stored in the stack, where we should save the codec? Always in the CODEC table or in stack too? 
If save in the stack I think we would have some problem with statements that look ups for codec from CODEC table, but on the other hand we will have an element not present in the HEAP which codec is present in the CODEC.

Moreover a look on the `then` part is needed if pointer is saved in the stack since we can print also the pointer.

jaromil:
We can soon revisit this and maybe change its purpose as I feel that the pointer function can be particularly useful to have in the Given phase, even before specific type definition, like import a dictionary,move pointer inside it and then import particular schema or types at pointer.

matteo-cristino:
Mh in the given phase we already have [Given '' in path ''](https://github.com/dyne/Zenroom/blob/master/src/lua/zencode_given.lua#L439-L445) for this puropose, no?

jaromil (approved):

PR #756, opened 2023-11-03
Feat: scope=given limited execution for schema introspection

jaromil:
This new feature adds the possibility to execute zencode limited to the Given and Then section. It also switches off the fatal condition for missing data imported by given, allowing the caller to ignore missing data and obtain a CODEC (schema).

The use-case is that of host applications willing to introspect the data schema accepted by a Zencode contract.


puria (approved):
It is 😍 just tested..

albertolerda (approved):

puria:
@jaromil tested now with a string dict and the codec is not printed... says
`[!] [source 'zencode_data']:226: Input encoding not found: string_dictionary`
same with `string array`, `number array` and so on

jaromil:
@puria yes, known bug still WIP. A full list of things left TODO is at bottom of test/zencode/given.bats

PR #762, opened 2023-11-13
alternative P256 approach with source inclusion

jaromil:
After noticing the architecture of p256-m this approach leverages it as a class in Lua space (P256) using octet as base type and implementing the same base API signature of other 1st order curves.


jaromil:
well done!!

PR #766, opened 2023-11-22
fix: 🐛 found statements work also with floats 

### LINKED ISSUE #754: verify is found in
Zencode: 

```
Given that I have a 'string dictionary' named 'Information'

If I verify 'myNumber' is found in 'Information' 
then print string 'found'
endif

Then print data 
```

Data: 

```
{
	"Information": {
		"Metadata": "TransactionsBatchB6789",
		"Buyer": "John Doe",
		"myNumber": 3
	}
}

```

Error: 

1698837621603W[W] Given that I have a 'string dictionary' named 'Information'
1698837621603W[W] D ZEN:pick found Information(d)
1698837621603W[W] If I verify 'myNumber' is found in 'Information'
1698837621603W[W] [!] /zencode_when.lua:40: attempt to get length of a zenroom.float value (field '?')
1698837621603E[!] ERROR:
1698837621603E[!] EXEC: /zencode.lua:803: If I verify 'myNumber' is found in 'Information'
1698837621603E[!] Execution aborted
1698837621603E[!] Execution aborted with errors.

matteo-cristino:
fix #754

### DIFF
--- a/src/lua/zencode_when.lua
+++ b/src/lua/zencode_when.lua
@@ -27,7 +27,7 @@ When("done", function() end)
 
 
 local function _is_found(el)
-    return ACK[el] and (luatype(ACK[el]) == 'table' or #ACK[el] ~= 0)
+    return ACK[el] and (luatype(ACK[el]) == 'table' or tonumber(ACK[el]) or #ACK[el] ~= 0)
 end
 
 IfWhen("verify '' is found", function(el)
--- a/src/lua/zencode_table.lua
+++ b/src/lua/zencode_table.lua
@@ -157,7 +157,7 @@ local function _is_found_in(ele_name, obj_name)
         end
     elseif obj_codec.zentype == 'd' then
         local el = O.to_string(ele)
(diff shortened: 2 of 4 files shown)


src/lua/zencode_when.lua
```diff
@@ -27,7 +27,7 @@ When("done", function() end)
 
 
 local function _is_found(el)
-    return ACK[el] and (luatype(ACK[el]) == 'table' or #ACK[el] ~= 0)
+    return ACK[el] and (luatype(ACK[el]) == 'table' or tonumber(ACK[el]) or #ACK[el] ~= 0)
```
jaromil, line 30: we have the safer `table_size()` in place of the `#` overload which may be convenient to use here.

matteo-cristino: The # is used to check that `ACK[el]` is not an empty string. Empty table are accepted since it return true at the first check and thus the others checks are not performed. Do we want to modify it and return as not found also empty tables?

jaromil:
xcellent!

PR #770, opened 2023-11-27
SD-JWT scenario

jaromil:
Implementation of Selective Disclosures in Zencode


andrea-dintino:
I think that the name of the "p256" keys should be changed to: 

sk: "ecdsa P-256"
pk: "ecdsa P-256 public key" 

We can discuss if the "P-256" should become "p256" for brevity, but the official name (see [NIST](https://csrc.nist.gov/csrc/media/events/workshop-on-elliptic-curve-cryptography-standards/documents/papers/session6-adalier-mehmet.pdf )) and also the name used in the SD-JWT standard, in "crv": "P-256" see [ietf](https://datatracker.ietf.org/doc/draft-ietf-oauth-selective-disclosure-jwt/).

PR #774, opened 2023-11-30
Feat/introspection

jaromil:
Improve the introspection capabilities of Zenroom to list all statements, tokens and stats connected, as well cleanup the init to narrow down strictly necessary scenarios. At last implement Norvig's autocorrection on detected tokens.

### DIFF
--- a/bindings/javascript/src/index.spec.ts
+++ b/bindings/javascript/src/index.spec.ts
@@ -535,10 +535,20 @@ test("decode zencode error", async (t) => {
 
 test("get all the statements", async (t) => {
   const res = await zencode_get_statements();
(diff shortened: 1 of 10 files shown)


matteo-cristino:
Looking at this PR I got an idea, can we also create an API that list all statements divided by `Given`, `When`, `Then`, `If` and `Foreach` (maybe also all Rules) and then wrap it into js? This will help to automatize the autocompletion for the editor we are working on. Any thoughts on this?

puria:
@matteo-cristino 😘

matteo-cristino:
Add the possibility to pass a scenario to zencode_get_statements

matteo-cristino (approved):

PR #779, opened 2023-12-11
docs: zencode sd-jwt

RebeccaSelvaggini:
SD-JWT documentation for Presentation and Verification phase, and for the final part of Credential Issuance flow (create SD-request, create signed-SD and disclosures).


jaromil (approved):
excellent!

PR #780, opened 2023-12-13
feat: ✨ numbers are always time or float independently from encoding

matteo-cristino:
expand also rules documentation


andrea-dintino:
merged by Andrea while sitting next to Jaromil

PR #784, opened 2023-12-19
feat(credential): ✨ add possiblity to import and export issuer public key in compressed form

matteo-cristino:


jaromil:
well done!!

Issue #288, opened 2022-01-24
Array of dictionaries


albertolerda:
If I have an array of dictionaries in the data and I declare it as 'string array' or 'string dictionary' no error is thrown.
To reproduce the problem, go to `apiroom.net`, use as data
```
{
	"query_result": [
		{
			"value": 300,
			"color": "blue"
		},
		{
			"value": "1300",
			"color": "yellow"
		}
	]
}
```
and with the script
```
Given I have a 'string dictionary' named 'query_result'
Then print data
```

andrea-dintino:
I found that using 'string array' works, the Zencode:

```
Given I have a 'string array' named 'array'
Given I have a 'number' named 'number'
Given I have a 'string' named 'string'
Given I have a 'string dictionary' named 'string dictionary'

When I insert 'number' in 'array'
When I insert 'string' in 'array'
When I insert 'string dictionary' in 'array'

Then print data

```

with data:

```
{
	"array": [
		{
			"name": "mario",
			"age": 2,
			"timestamp": 3
		},
		{
			"name": "luigi",
			"age": 5,
			"timestamp": 2
		}
	],
	"number": 128,
	"string": "my sexy string",
	"string dictionary": {
		"ip": "0.0.0.1",
		"pubkey": "abcdesfdsdsdfd",
		"ciccio": "pesce"
	}
}
```

@jaromil maybe we can create 'dictionary array' as an alias of 'string array'? 

Produces the output: 

```
{
  "array": [
    {
      "age": 2,
      "name": "mario",
      "timestamp": 3
    },
    {
      "age": 5,
      "name": "luigi",
      "timestamp": 2
    },
    128,
    "my sexy string",
    {
      "ciccio": "pesce",
      "ip": "0.0.0.1",
      "pubkey": "abcdesfdsdsdfd"
    }
  ]
}
```

jaromil:
The name array or dictionary must be in second position to the definition of encoding when declared.

The definitions array or dictionary refer to the 1st level of depth of the object. So it is correct to use '*encoding* array'

matteo-cristino:
fixed in #788

PR #788, opened 2023-12-21
fix(given): 🐛 add data type validation during giving phase

matteo-cristino:
This will help to ensure that the zentype declared in the codec is the right one and that the statements which act differently based on the zentype works as intended.


matteo-cristino:
fix #288

PR #811, opened 2024-02-26
feat(w3c): ✨ more generic jws creation and verification

matteo-cristino:
Implemented statements are:
* `When I create jws signature with header '' payload '' and signature ''`
*  `When I verify jws signature in ''`

The second one is similar to the already implemented `When I verify jws signature of ''`, but the latter takes in input the data that has been signed in the jws (to be used when payload is not present), while the above one takes in input the jws itself where what we signed is the payload present in the jws.

jws form:
`header`.`payload`.`signature`


matteo-cristino:
Deprecated statements:
* create jws signature of ''
* create jws signature using ecdh signature in ''
* verify jws signature of ''

New statements:
* create jws signature with header '' and payload ''
* create jws detached signature with header '' and payload ''
* verify '' has a jws signature in ''
* verify jws signature in ''

Improvements in jws:
* check header and payload are always json (encoded or not)
* sign always header.payload
* for retrocompatibility verify also raw payload signature (but a warn in this case is returned)
* signature and verification are autodetected from header (no need to specify keys or scenarios)

src/lua/zencode_w3c.lua
```diff
@@ -145,85 +145,228 @@ ZEN:add_schema(
...
+-- @return public key, tsake the one in the header (if present) or the correct key in zenroom memory
+-- @return header.payload, if payload is differenet from the empty string
+local function jws_octet_to_signature(o_jws, o_payload)
+    local toks = strtok(OCTET.to_string(o_jws), '.')
+    -- parse header
+    local pk
+    local header = JSON.decode( OCTET.from_url64(toks[1]):string())
+    if header.jwk then
+        -- kty MUST be present
+        zencode_assert(header.jwk.kty == 'EC',
```
matteo-cristino, line 166: use error('',2) instead of zencode_assert in jws_octet_to_signature

src/lua/zencode_w3c.lua
```diff
@@ -145,85 +145,228 @@ ZEN:add_schema(
...
+                       'JWS public key type supported by zenroom is only EC')
+        -- if kty == EC, crv MUST be present
+        zencode_assert(header.jwk.crv == 'P-256' or header.jwk.crv == 'secp256k1',
+                       'JWS public key curve supported by zenroom are only P-256 or secp256k1')
+        pk = O.from_url64(header.jwk.x) .. O.from_url64(header.jwk.y)
+    end
+    zencode_assert(header.alg, 'JWS header is missing alg specification')
```
matteo-cristino, line 173: check that alg matches jwk.crv in the case there is the jwk key

src/lua/zencode_w3c.lua
```diff
@@ -145,85 +145,228 @@ ZEN:add_schema(
...
+    local o_payload = _json_encoding(payload)
+    empty(n_output)
+    ACK[n_output] = O.from_string(
+        jws_signature_to_octet(nil, o_header, o_payload, detached)
+    )
+    new_codec(n_output, { zentype = 'e',
+                          encoding = 'string' })
+end
+
+When("create jws signature with header '' and payload ''", function(header, payload)
```
matteo-cristino, line 308: create the jws signature of header '' and payload ''

src/lua/zencode_w3c.lua
```diff
@@ -145,85 +145,228 @@ ZEN:add_schema(
...
+                "create jws detached signature with header '' and payload ''",
+                function(sign)
+                    warn('external signature can be not jws compliant')
+                    local signature = have(sign)
+                    empty'jws'
+                    ACK.jws = O.from_string(jws_signature_to_octet(signature))
+                    new_codec('jws', { zentype = 'e',
+                                       encoding = 'string' })
+end))
+
```
matteo-cristino, line 295: add stataments to create jws header like:
* create jws header for es256 signature 
* create jws header for es256 signature with public key (if es256 key present, compute pk and add to header as jwk)

both for es256/secp256r1/p256 and ecdh/es256k/secp256k1

PR #822, opened 2024-03-05
feat: ✨ various improvements on move and copy statements

matteo-cristino:


jaromil:
Perhaps we can also pack a new statement into this PR:

```
When I create the random 'ramdom_jti'
When I write the string '' in 'jti'
When I append the 'url64' of 'ramdom_jti' to 'jti'
```

We can have a way to `copy` and `move` as `url64` and other encodings inside a string dictionary:
```
When I move the `random_jti` as `url64` in `jti`
```

puria:
If I may suggest taking inspiration from HyperTalk https://en.wikipedia.org/wiki/HyperTalk (which unfortunately no longer exists, but its syntax has always fascinated me and its derivatives are making a comeback). 
In my humble opinion, the phrases **move** and **copy** (and the use of **in**) could be improved.
I don't want to explicitly suggest... but take a look at it. 

Incidentally, xTalk/HyperTalk it was also used for `script poems` or `Script as Poetry` (which closely aligns with the Dyne.org philosophy) and the _porno recursion_ is quite famous: https://www2.iath.virginia.edu/elab/hfl0128.html

PR #828, opened 2024-03-11
feat(when): ✨ statement to exit with error message

matteo-cristino:


jaromil:
Nice! we meant to have it since quite some time now.

PR #829, opened 2024-03-12
feat(http): ✨ create get parameters string from a dictionary of parameters

matteo-cristino:
I'm not so sure about the statements syntax 🤔


src/lua/zencode_http.lua
```diff
@@ -163,3 +163,28 @@ end)
...
+    end
+    local res = ""
+    for k,v in pairs(params) do
+        res = res..encoding_f(k).."="..encoding_f(v:str())
+    end
+    ACK[dest] = O.from_string(res)
+    new_codec(dest, { zentype = 'e', encoding = 'string' })
+end
+
+When("create get parameters from ''", function(table_params)
```
jaromil, line 184: perhaps better:

`create http get parameters from''`

`create http get parameters from'' using percent encoding`

PR #831, opened 2024-03-13
refactor(sd_jwt): updated supported_selective_disclosure

RebeccaSelvaggini:
Now supported_selective_disclosure is imported/exported and used as the new credential_issuer well-known.


src/lua/zencode_sd_jwt.lua
```diff
@@ -27,110 +27,94 @@ local function import_url_f(obj)
...
+               if v == needed[k] then
+                    found = true
+                    break
+               end 
             end
+        else 
+            found = obj[what] == needed
         end
-        assert(found, needed .. " not supported in " .. what)
+        assert(found, " not supported in " .. what)
```
jaromil, line 72: since this is not inside a zencode statement function, the error(..., 2) call should be used instead with a check, so that the line returned by the error will be pointing to the caller of this function.

src/lua/zencode_sd_jwt.lua
```diff
@@ -27,110 +27,94 @@ local function import_url_f(obj)
...
-
-    local creds = obj.credentials_supported
+    res.authorization_servers = schema_get(obj, 'authorization_servers', import_url_f, tostring)
+    local creds = obj.credential_configurations_supported
     for i=1,#creds do
         check_display(creds[i].display)
-        local count = 0
-        for j=1,#creds[i].order do
-            local display = creds[i].credentialSubject[creds[i].order[j]]
+        check_support(creds[i], 'format', 'vc+sd-jwt')
```
jaromil, line 84: check_support error depth is 3 since its called from here.

jaromil:
well done!

PR #840, opened 2024-04-03
feat(sd_jwt): ✨ statement to verify if a set of disclosures is contained in a sd_jwt

matteo-cristino:


matteo-cristino:
@jaromil I'm not so sure about the "contains" word, any idea on that?

src/lua/zencode_sd_jwt.lua
```diff
@@ -416,6 +416,21 @@ When("use signed selective disclosure '' only with disclosures ''", function(ssd
     ssd.disclosures = disclosure
 end)
 
+IfWhen("verify selective disclosure '' contains disclosures ''", function(ssd_name, lis)
```
jaromil, line 419: more aligned with other statements to use:

`is found` or `are found`

`verify disclosures '' are found in selective disclosure ''`

perhaps we can make `is` and `are` equivalent

src/lua/zencode_sd_jwt.lua
```diff
@@ -416,6 +416,21 @@ When("use signed selective disclosure '' only with disclosures ''", function(ssd
...
+    local ssd = have(ssd_name)
+    local disclosed_keys = have(lis)
+    for _,k in pairs(disclosed_keys) do
```
jaromil, line 422: check that this is an array else pairs will act weird

also accept case in which there is a single one being verified fo be found.

jaromil (changes requested):

src/lua/zencode_sd_jwt.lua
matteo-cristino: Used `are found`

src/lua/zencode_sd_jwt.lua
matteo-cristino: added checks and possibility to have a single element

jaromil (approved):

jaromil:
Well done. Later on we may want to  `gsub(tt, ' are ', ' is ')` in the zencode parser, but will deserve another PR.

PR #841, opened 2024-04-03
feat(foreach): ✨ add statement to exit from a foreach loop

matteo-cristino:


src/lua/zencode_foreach.lua
```diff
@@ -64,3 +64,8 @@ Foreach("'' in sequence from '' to '' with step ''", function(name, from_name, t
         end
     end
 end)
+
+When("exit from foreach loop", function()
```
jaromil, line 68: best use:

'exit foreach loop`

which will support also `exit the`

unsure if to use `break` as a verb, more used in other langs

matteo-cristino: If we are unsure qe can have them as alias both `exit` and `break`

jaromil: Yes to aliasing both `exit` and `break`.

I checked and the word `loop` is not mentioned anywhere so:

```
break foreach
exit foreach
```

One doubt left since `endforeach` is used as one word to mark end of branch... I think there will be confusion later on about having the word separated, as well between `exit` and `end`.

jaromil:
I think my doubts can be set aside and as you say it seems coherent to break inside a statement rather than prefix. Merging. Very well done!

PR #845, opened 2024-04-04
deterministic sort of dictionaries using `zenroom.` types as keys

### LINKED ISSUE #838: deterministic sort_pairs cannot sort tables with `zenroom.` type keys
The adoption of vanilla `table.sort` from lua here: https://github.com/dyne/Zenroom/blob/6262bc2904e60b68a463740f9516feb6b2c45d32/src/lua/zenroom_common.lua#L110

Breaks the deterministic sorting of tables in case they are dictionaries and the keys are a `zenroom.` type.

We need to implement our own sort to support also zenroom types (it can be over octet bytes).

jaromil:
For later on, it would be handy to be able to use `zenroom.` type keys in dictionaries and have them sorted deterministically: this would ease for instance filtering of selective disclosure in which octets of hashes are used as keys and signatures as values. There are other use-cases for this. 

fix #838


jaromil:
https://github.com/dyne/Zenroom/issues/838

PR #847, opened 2024-04-05
feat: ✨ add support for time on explicit sum operations in zencode

matteo-cristino:


jaromil:
This seems very well done. I think that implementing subtraction should be enough to finish the PR, since other operations are less likely to be needed on `time` elements. Also we need to implement comparisons (equal, less or more than)

jaromil:
Awesome!

PR #865, opened 2024-04-30
chore: update emsdk

puria:


jaromil:
just out of curiosity, can we know the difference in time of execution between js tests with this branch and current HEAD?

jaromil:
Can we activate the emsdk mimalloc port?
from changelog:
Add a port of mimalloc, a fast and scalable multithreaded allocator. To use it, build with -sMALLOC=mimalloc. (#20651)

jaromil:
Check how this affects us (from 3.1.36 chlog):/

The following JavaScript runtime functions were converted to JavaScript library functions:

UTF8ArrayToString
UTF8ToString
stringToUTF8Array
stringToUTF8
lengthBytesUTF8
If you use any of these functions in your JS code you will now need to include them explicitly in one of the following ways:

Add them to a __deps entry in your JS library file (with leading $)
Add them to DEFAULT_LIBRARY_FUNCS_TO_INCLUDE (with leading $)
Add them to EXPORTED_FUNCTIONS (without leading $)
Set -sLEGACY_RUNTIME to include all of them at once.

Also consider using `-sSTRICT`

.github/workflows/main.yml
```diff
@@ -340,7 +340,7 @@ jobs:
     if: ${{ needs.semantic-release.outputs.release == 'True' }}
     runs-on: ubuntu-latest
     env:
-      EM_VERSION: 3.1.26
```
matteo-cristino, line 343: maybe this can be set to latest as in the test job? Or there was a particular reason to stick with this specific version?

jaromil: ah no ok I see what you mean. Yes it is so that it doesn't breaks when we have no time to review the upgrade.

puria: Fort testing is going to be okay. For release we should have a way to reproduce builds in case of bugs. You don't want to release zenroom-js version without knowing the exact build tools. Nice observation, though!

PR #868, opened 2024-05-02
RSA signature support

jaromil:
Here we add support for RSA 2048 and 4096 signatures with a simple interface extending milagro's primitives.

For the API reference see `rsa_*.h` in `build/milagro/include` and `rsa_support.h/c` in `lib/milagro../src`


matteo-cristino:
I see some discrepancy in the tabs in zen_rsa.c and zencode_rsa.lua (spaces vs tabs). Can you fix it before merging so that we have not to deal with it later on?

PR #879, opened 2024-05-13
feat: ✨ add zencode contract parse api

matteo-cristino:
- feat: add strict_parse configuration flag
- fix: return of parse is json encoded
- fix: parser does not ignore invalid statement that are not at the contract borders
- feat(zenroom): ✨ add new C api to parse zencode contracts
- feat(javascript): ✨ expose zencode_parse_contract api to js
- refactor: ♻️  add parse api input to decide whether it has to be strict or not
- feat(javscirpt): ✨ parse api is by default strict, add also safe_zencode_parse_contract as an alias for non strict parsing

### DIFF
--- a/bindings/javascript/src/index.spec.ts
+++ b/bindings/javascript/src/index.spec.ts
@@ -9,6 +9,7 @@ import {
   zenroom_hash,
   introspect,
   zencode_parse_contract,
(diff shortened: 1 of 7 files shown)


jaromil (commented):
well done!! just need help to re-align js wrapper signature for the new api call

PR #882, opened 2024-05-13
fix: 🐛 print empty octets to string as empty strings

matteo-cristino:
- fix: print empy octets to string as empty strings
- test(zencode): add test for import and export of empty strings in tables


jaromil:
good catch!!!

PR #883, opened 2024-05-13
fix: print all empty octets as empty strings

matteo-cristino:
- fix: print all empty octets as empty strings
- test(zencode): add failing tests with empty octets in the trace

### DIFF
--- a/test/zencode/zencode_exec.bats
+++ b/test/zencode/zencode_exec.bats
@@ -196,25 +196,23 @@ EOF
 	assert_output '{"myMessage":"Dear Bob, your name is too short, goodbye - Alice.","output":["Zenroom_certifies_that_signatures_are_all_correct!"]}'
 }
 
-# TODO: how to test failure and error message with zencode_exec?
 @test "failing test print encoded trace with empty string in it" {
-    skip
 
-	# empty conf
-	echo > zencode_exec_stdin
+    # empty conf
+    echo > zencode_exec_stdin
 
     # zencode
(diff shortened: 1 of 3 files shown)


matteo-cristino:
@jaromil I think I will need your help for this part https://github.com/dyne/Zenroom/pull/883/files#diff-2d5670cdc813fe0e61b94eb79c029066d90b15a37ddffc012283322702c3920dR199, not quite sure of how to do it

test/zencode/zencode_exec.bats
```diff
@@ -195,3 +195,44 @@ EOF
 	save_output verified.json
 	assert_output '{"myMessage":"Dear Bob, your name is too short, goodbye - Alice.","output":["Zenroom_certifies_that_signatures_are_all_correct!"]}'
 }
+
+# TODO: how to test failure and error message with zencode_exec?
```
jaromil, line 199: zencode-exec returns Z->exitcode whose values are defined in zenroom.h
```c
// EXIT CODES
#define ERR_INIT 4
#define ERR_PARSE 3
#define ERR_EXEC 2
#define ERR_GENERIC 1 // EXIT_FAILURE
#define SUCCESS 0 // EXIT_SUCCESS
```

matteo-cristino: At the end I was able to catch the message error! Bats seems not to save the output and error when using a `cat file | run executable`

jaromil:
many thanks for this additional test!

PR #886, opened 2024-05-23
Docs/ml kem 512

ManudL2000:
We added to the documentation one example of the Key Encapsulation Mechanism ml-kem-512.


matteo-cristino:
Just some small things, can you please update the zencode mlkem script with
```
Rule check version 4.32.0
```
instead of
```
Rule check version 2.0.0 
```
since this feature is available only from this version. Moreover can you eliminate all files under the path `docs/examples/zencode_cookbook/mlkem512/` that are not used in the documentation? To have a clear look at what is used in the documentation and what not in case of future update.

Grate work 🚀

PR #888, opened 2024-05-26
fix(zen_ecp): 🐛 zen_arg must return elements allocated from C and arg_free must only and always be used after it

matteo-cristino:


jaromil:
thanks! I stand corrected on previous PR, well spotted.

PR #892, opened 2024-06-24
fix(zenroom): initialize zenroom before script check, in this way the error on empty script follows the logfmt specification

matteo-cristino:
- fix(zenroom): initialize zenroom before script check, in this way the error on empty script follows the logfmt specification
- test(js): update error returned by empty script


matteo-cristino:
There are still some `_err()` calls that does not follow the logfmt specified in the conf, but these can not be directly substituted with `zerror()` since are zenroom or lua initialization error. Another solution should be found for these cases to report errors with the right format.

puria (approved):

PR #898, opened 2024-07-05
fix: increment js binding wasm stack size to 7MB and correct reported error from lerror

matteo-cristino:
- fix(js): increment wasm stack size to 7MB
- test(js): zenroom correctly fails on input larger than 4MiB
- fix: result octect length computation when decoding from base64
- chore(js): update huge_input file to really have a varibale whose size is bigger than the accepted one
- test(js): update test to new input
- fix(zen_error): lerror does not forward optional args to zerror
- chore: remove some useless printing from test file

### DIFF
--- a/bindings/javascript/src/index.spec.ts
+++ b/bindings/javascript/src/index.spec.ts
@@ -1,5 +1,4 @@
 import test from "ava";
-import fs from "node:fs/promises";
 
 import {
   zencode_exec,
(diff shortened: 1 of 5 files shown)


puria (approved):

jaromil:
can we generate the huge json file instead of committing it inside the git repo?

matteo-cristino:
You are right, I think we can generate it with something like
```
{"keyring": {"ecdh": "kERUCRJekQrsKsfvNVnw3K7Iw4WnhCvGbHqsMeD8arE="}},"bytes": ${'a'.repeat(5*1024*1024)}}
```
Will check it tomorrow morning

jaromil:
good catch on lerror() !!

jaromil:
I will merge this after removing the 7mb file from the history, it may be necessary a squash or preferably merging from a new branch to preserve history if single commits.

matteo-cristino:
What about squashing directly this PR on merging?

jaromil:
ok, merged manually with a squash. well done!!

PR #901, opened 2024-07-17
Fuzzing methods

jaromil:
Add some handy functions to flip bits and bytes in random positions inside octets, or at beginning or end, using xor or random values.


matteo-cristino:
Please pay attention to indentation, in C it is tabs 8 spaces long

Issue #866, opened 2024-05-02
Small heap buffer overflow in ecdh_pubgen


jaromil:
Bug found running zencode tests with a `linux-asan` build

## Test unit:
```
keys.bats
 ✓ Generate key seed
 ⚙️ generate-key-seed.zen
   📝 generate-key-seed.keys
 💾 generate-key-seed.out

 ✗ String seed
 ⚙️ generate-key-seed-string.zen
   📝 string-seed.json
```
## Details
```
 ==138149==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60300002db3a at pc 0x562ecf10eeb0 bp 0x7ffc6fcfd520 sp 0x7ffc6fcfd510
   READ of size 1 at 0x60300002db3a thread T0
       #0 0x562ecf10eeaf in BIG_256_28_fromBytes /home/jrml/devel/zenroom/lib/milagro-crypto-c/build/src/big_256_28.c:162
       #1 0x562ecf11c6f6 in ECP_SECP256K1_KEY_PAIR_GENERATE /home/jrml/devel/zenroom/lib/milagro-crypto-c/build/src/ecdh_SECP256K1.c:49
       #2 0x562ecf038d30 in ecdh_pubgen /home/jrml/devel/zenroom/src/zen_ecdh.c:182
```
```
 0x60300002db3a is located 0 bytes to the right of 26-byte region [0x60300002db20,0x60300002db3a)
   allocated by thread T0 here:
       #0 0x7f24208c9887 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
       #1 0x562ecf00f4b0 in o_new /home/jrml/devel/zenroom/src/zen_octet.c:222
       #2 0x562ecf00ff5f in o_dup /home/jrml/devel/zenroom/src/zen_octet.c:329
       #3 0x562ecf038cce in ecdh_pubgen /home/jrml/devel/zenroom/src/zen_ecdh.c:170
```

PR #902, opened 2024-07-17
fix(zen_ecdh): check secret key size on generation of the public one

matteo-cristino:
- fix(zen_ecdh): check secret key size on generation of the public one
- chore(zencode_ecdh): avoid code duplication


matteo-cristino:
fix #866

PR #905, opened 2024-07-23
feat(time): date table to and from timestamp conversions and math operations

matteo-cristino:


matteo-cristino:
Need an help to craft a better statement, the one that I wrote is not so clear in my opinion ahahah

Moreover not so sure on how to test it 🤔

jaromil:
I'm in favor of merging this, can switch to ready any time if you agree @matteo-cristino

PR #909, opened 2024-07-26
Improve bbs perfomance 

FilippoTrotter:
Moved the implementation of map_to_curve from Lua to C, and now the tests are almost 70% faster.


jaromil:
Promising benchmarks, using a 12th gen i9 cpu

keygen
sign_verify

ZKP benchmarks increase the number of credentials and disclose half of them at each step
zkp

matteo-cristino:
Before merging can you have a look at the indentation of zen_bbs.c? Since it is a new file also here we should start with the correct indentation (for C files are tabs long 8 spaces)

PR #912, opened 2024-07-31
fix rsa keygen docs script

FilippoTrotter:


andrea-dintino (commented):
One more issue, in the verification script, to load the RSA pubkey you should use:

Given I have a 'rsa public key' in 'Alice'

@FilippoTrotter

PR #924, opened 2024-09-17
For docker image FROM python:3.9 need xxd also

Ignat99:
Changes in ReadMe file about Dependencies from xxd

I use docker image FROM python:3.9
in order to install the package via github, because I think to add additional code to one lua file of the project and so that the changes are available via Python 3.8+ module zenroom

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ ready to deploy.
 
 ## Build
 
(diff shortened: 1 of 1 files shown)


jaromil:
We previously found out that `xxd` is sometimes present as package, sometimes shipped with vim package. It's OK to add this, but so you know why it wasn't mentioned.

PR #926, opened 2024-09-19
ethwalletbypassword 

Ignat99:
Get wallets fpom miner by password in private blockchain

I use [Private blockchain](https://medium.com/scb-digital/running-a-private-ethereum-blockchain-using-docker-589c8e6a4fe8)

All access is password only. I usually don't have access to the private key.

In this addendum to the documentation I propose 2 methods to solve this issue.

One is by directly reading the keyring file.

The second is [more or less the standard way](https://github.com/ethereum/go-ethereum/blob/master/cmd/clef/tutorial.md) of remote access via clef. In this case we can use an improvised backend based on one clef option

`Until then however, we're trying to pave the way via Geth. Geth v1.9.0 has built in support via --signer <API endpoint> for using a local or remote Clef instance as an account backend!`


Ignat99:
Do you think it is possible to add this method to Zencode?

When we only pass the path to the keyring file and the password.
And from this data we generate a wallet and, for example, a public key, as is customary in examples.

Because without this property, all the examples seemed educational to me, which did not solve my problem.

This is mainly useful for beginners who have not yet gotten their bearings in the system.

But if we want people to use smart contracts and Zenroom, then we should make the information more accessible, and the examples closer to the real task.

jaromil:
Looks good to me! indeed its an easy example to get the private key imported in zenroom.

PR #928, opened 2024-09-23
ci: update docker release

matteo-cristino:
CI runs only on tags push

### DIFF
--- a/.github/workflows/publish-docker.yml
+++ b/.github/workflows/publish-docker.yml
@@ -2,63 +2,12 @@ name: Create and publish a Docker image
 
 on:
   push:
-    branches:
-      - 'master'
     tags:
       - 'v*'
-  pull_request:
-    branches:
-      - 'master'
-
-env:
-  REGISTRY: ghcr.io
-  IMAGE_NAME: dyne/zenroom
 
 jobs:
-  build-and-push-image:
-    runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      packages: write
-
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v3
-
-      - name: Extract metadata (tags, labels) for Docker
-        id: meta
-        uses: docker/metadata-action@v3
-        with:
-          images: |
-            ${{ env.IMAGE_NAME }}
-            ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
-          tags: |
-            type=edge
-            type=sha
-            type=semver,pattern={{version}}
-            type=semver,pattern={{major}}.{{minor}}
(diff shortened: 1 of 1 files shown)


Ignat99:
It would probably make sense to create a docker-compose file with all images, including [Planetmint](https://github.com/planetmint) and [Zenswarm Storage](https://forkbomb.solutions/component/zenswarm-storage/)?

jaromil:
> It would probably make sense to create a docker-compose file with all images, including [Planetmint](https://github.com/planetmint) and [Zenswarm Storage](https://forkbomb.solutions/component/zenswarm-storage/) and [Fabchain](https://github.com/dyne/fabchain)?

These two projects are discontinued, as they didn't gain any traction among users. We only maintain generic integration now.

> Because tests need to be done on signed transactions and tokens (ERC 721, ERC 1155) with ready-made smart contracts. Unfortunately, I have not yet been able to find a ready-made example of using [these features](https://github.com/dyne/Zenroom/blob/master/test/ethereum/transfer_eth_plt.sh).

> The example above provides a ready-made contract without any explanation of what it looks like, where it is located, and how to assemble your own distributed data storage.

We have a ganache server running on a VPS which is used to test against. If you like to help us maintain it, that would be great. Its running as-is since a couple years with the contracts we test.

Ignat99:
I have a docker-compose file for ganache with a tunnel via ssh for a remote graphical desktop. But I'm thinking about how to do it easier and more standard. But since I'm not an expert in Ethereum distributed databases specifically, I'm only studying how this is done standardly in Ethereum. Do I need an additional image for the data storage in addition to the existing boot node, RPC and miner?

Ignat99:
> > It would probably make sense to create a docker-compose file with all images, including [Planetmint](https://github.com/planetmint) and [Zenswarm Storage](https://forkbomb.solutions/component/zenswarm-storage/) and [Fabchain](https://github.com/dyne/fabchain)?
> 
> These two projects are discontinued, as they didn't gain any traction among users. We only maintain generic integration now.
> 
> > Because tests need to be done on signed transactions and tokens (ERC 721, ERC 1155) with ready-made smart contracts. Unfortunately, I have not yet been able to find a ready-made example of using [these features](https://github.com/dyne/Zenroom/blob/master/test/ethereum/transfer_eth_plt.sh).
> 
> > The example above provides a ready-made contract without any explanation of what it looks like, where it is located, and how to assemble your own distributed data storage.
> 
https://github.com/dyne/Zenroom/blob/master/test/ethereum/create_erc721.sh#L52C27-L52C35

Here I clearly see a request to the test network Fabchain.

PR #937, opened 2024-09-29
chore: remove yarn from javascript dependencies

matteo-cristino:
- chore: remove yarn from javascript dependencies
- ci(js): remove cwd since working directory is already set to the right folder


jaromil:
I've changed the way this PR works by keeping the workdir at root in CI and keeping --cwd

This way it will be aligned with the new makefile system which will also use --cwd

PR #949, opened 2024-10-14
BBS+: add some checks on disclosed indexes array

FilippoTrotter:


jaromil:
ahah!! so you read my email and blog !!!! <3 thanks a lot for taking the initiative !!! 👍🏽

src/lua/crypto_bbs.lua
```diff
@@ -644,16 +644,19 @@ local function core_proof_gen(ciphersuite, pk, signature, generators, header, ph
     local signature_result = octets_to_signature(signature)
     local AA, e = table.unpack(signature_result)
     local undisclosed_indexes = {}
+    local disclosed_messages = {}
+    local undisclosed_messages = {}
+    for i, v in ipairs(disclosed_indexes) do
+        if i > 1 and disclosed_indexes[i] == disclosed_indexes[i - 1] then
```
jaromil, line 650: I am not sure that checking only adjacent values suffices here. There can be scattered duplicates.

src/lua/crypto_bbs.lua
```diff
@@ -679,6 +682,7 @@ function bbs.proof_gen(ciphersuite, pk, signature, header, ph, messages, disclos
     header = header or O.empty()
     ph = ph or O.empty()
     table.sort(disclosed_indexes) -- make sure indexes are sorted
+    if (disclosed_indexes[1] <= 0 or disclosed_indexes[#disclosed_indexes] > #messages)  then error('disclosed indexes contains not valid integers', 2) end
```
jaromil, line 685: It is best to separate this check in two and give a more accurate error message describing the cause.

src/lua/crypto_bbs.lua
FilippoTrotter: Arrays are always sorted in proof_gen (line 684) before being passed as input to core_proof_gen

jaromil: can you add this sentence as comment in that line of code?

PR #951, opened 2024-10-15
fix(http): allow generic schema and to append numbers as query parameters

matteo-cristino:
- fix(http): improve uri parsing
- fix(http): allow to append also numbers as url query parameters


jaromil (approved):
Excellent

PR #952, opened 2024-10-15
feat(sd_jwt): create the key-value table of disclosed informations

matteo-cristino:
Utility that will simplify the code in the relying party.

I'm not so sure about the name `disclosed informations`, maybe there is a proper name for this object?

### DIFF
--- a/src/lua/zencode_sd_jwt.lua
+++ b/src/lua/zencode_sd_jwt.lua
@@ -478,3 +478,19 @@ IfWhen("verify signed selective disclosure '' issued by '' is valid", function(o
     end
 
 end)
+
+When("create disclosed informations from signed selective disclosure ''", function(ssd_name)
+    local ssd, ssd_c = have(ssd_name)
+    zencode_assert(ssd_c.schema and ssd_c.schema == "signed_selective_disclosure",
+        "Object is not a signed selective disclosure: " .. ssd_name)
+    local disclosed_informations = {}
+    for _, v in pairs(ssd.disclosures) do
+        disclosed_informations[v[2]:str()] = v[3]
+    end
+    ACK.disclosed_informations = disclosed_informations
+    new_codec('disclosed_informations', {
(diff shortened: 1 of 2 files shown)


jaromil:
I feel that we were mistaken in the first place to use the term `disclosure`, which should be attributed to what you intend here with disclosed data. Perhaps we should name hashes as `selective proof` and then use `selective disclosure` for clear text.

In any case, the best would be to ask also @andrea-dintino on what is commonly used as term in other wallets.

src/lua/zencode_sd_jwt.lua
```diff
@@ -478,3 +478,19 @@ IfWhen("verify signed selective disclosure '' issued by '' is valid", function(o
...
 
 end)
+
+When("create disclosed informations from signed selective disclosure ''", function(ssd_name)
+    local ssd, ssd_c = have(ssd_name)
+    zencode_assert(ssd_c.schema and ssd_c.schema == "signed_selective_disclosure",
+        "Object is not a signed selective disclosure: " .. ssd_name)
+    local disclosed_informations = {}
+    for _, v in pairs(ssd.disclosures) do
+        disclosed_informations[v[2]:str()] = v[3]
```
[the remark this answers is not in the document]
matteo-cristino: Salt, that is the first value in the array, is discarded, the key-value pair is saved into the new dictionary

src/lua/zencode_sd_jwt.lua
```diff
@@ -478,3 +478,19 @@ IfWhen("verify signed selective disclosure '' issued by '' is valid", function(o
     end
 
 end)
+
+When("create disclosed informations from signed selective disclosure ''", function(ssd_name)
```
jaromil, line 482: s/informations/information/ since there is no plural for that word in english.

matteo-cristino: Outdated

PR #953, opened 2024-10-18
Verify that SmartCondract realy deployed and work

Ignat99:
I noticed some shortcomings of the script when testing it for ERC 1155.

If you approve my intention to add to Zenroom/tree/master/docs/examples/zencode_cookbook/ethereum a minimal working contract for ERC 1155 and a script for testing token generation, I can make another PR.

But I would not like to overload this page with redundant Python scripts. They are only information for thinking about how best to do this using ZenCode.


ignat99inkan:
[https://ethereum.stackexchange.com/questions/150281/invalid-opcode-opcode-0x5f-not-defined](https://ethereum.stackexchange.com/questions/150281/invalid-opcode-opcode-0x5f-not-defined)

This is because solidity 0.8.20 introduces the PUSH0(0x5f) opcode which is only supported on ETH mainnet and not on any other chains. That's why other chains can't find the PUSH0(0x5f) opcode and throw this error.

To fix this error make sure solidity version is less that 0.8.20. I used 0.8.19 both in .sol file and package.json

Instead of going back to an older version of Solidity compiler, in hardhat to prevent use of PUSH0 you can try setting evmVersion to a previous version (e.g. paris, which is the one before shanghai) 

`solidity: {
    compilers: [
      {
        version: `0.8.21`,
        settings: {
          optimizer: {
            enabled: true,
            runs: 15000
          },
          evmVersion: `paris`
        }
      },
    ],
  },`

Ignat99:
For latest OpenZeppelin 5.0.0 - 5.1.0 version of Smart contract need add next settings.

[invalid opcode: opcode 0x5f not defined](https://ethereum.stackexchange.com/questions/150281/invalid-opcode-opcode-0x5f-not-defined)

This is because solidity 0.8.20 introduces the PUSH0(0x5f) opcode which is only supported on ETH mainnet and not on any other chains. That's why other chains can't find the PUSH0(0x5f) opcode and throw this error.

To fix this error make sure solidity version is less that 0.8.20. I used 0.8.19 both in .sol file and package.json

Instead of going back to an older version of Solidity compiler, in hardhat to prevent use of PUSH0 you can try setting evmVersion to a previous version (e.g. paris, which is the one before shanghai) 

```javascript
solidity: {
    compilers: [
      {
        version: `0.8.21`,
        settings: {
          optimizer: {
            enabled: true,
            runs: 15000
          },
          evmVersion: `paris`
        }
      },
    ],
  },
```

jaromil:
Hey many thanks! I think this improves and completes well the documentation on ethereum in Zenroom. You are welcome to move forward with the next PR.

PR #954, opened 2024-10-22
ci: add arm32 build test

matteo-cristino:
This reproduces @andrea-dintino error trying to build zenroom on arm32 with `make -f build/linux.mk` command


jaromil:
I guess it is mimalloc being problematic, would be enough to switch it off. Will take into account in the new build and rebase this for test. Thanks!!

matteo-cristino:
A quick solution that does not require to switch off mimalloc can be also to link the libatomic library with `-latomic` flag at build time.

PR #955, opened 2024-10-23
fix(http): separate hostname and port before host checks

matteo-cristino:
- fix(http): separate hostname and port before host checks
- test(http): parse url with port


jaromil:
looks ok! _is_valid_host used now only in one place but makes sense to keep as function for readability

PR #965, opened 2024-11-16
fix: minus sign in parathesis

matteo-cristino:
first minus sign in paranthesis was always considered as unary minus also in case like (a-b) leading to errors


jaromil:
wow tricky one... well done!!!!

PR #972, opened 2024-11-21
docs: update js documentation and fix react example

matteo-cristino:
- docs: update some js documentation
- docs: fix react documentation

### DIFF
--- a/docs/pages/javascript.md
+++ b/docs/pages/javascript.md
@@ -28,20 +28,36 @@
 Stable releases are published on https://www.npmjs.com/package/zenroom that
 have a slow pace release schedule that you can install with
 
+<!-- tabs:start -->
+
+### ** npm **
+
 ```bash
-yarn add zenroom
-# or if you use npm
 npm install zenroom
 ```
 
+### ** yarn **
+
+```bash
+yarn add zenroom
+```
+
+### ** pnpm **
+
+```bash
+pnpm add zenroom
+```
+
+<!-- tabs:end -->
 
 * * *
 
 ## 🎮 Usage
 
-The binding consists of one main function::
+The binding consists of two main functions:
 
-**zencode_exec** to execute [Zencode](https://dev.zenroom.org/#/pages/zencode-intro?id=smart-contracts-in-human-language). To learn more about zencode syntax look [here](https://dev.zenroom.org/#/pages/zencode-cookbook-intro)
+- **zencode_exec** to execute [Zencode](https://dev.zenroom.org/#/pages/zencode-intro?id=smart-contracts-in-human-language). To learn more about zencode syntax look [here](https://dev.zenroom.org/#/pages/zencode-cookbook-intro)
+- **zenroom_exec** to execute our special flavor of Lua enhanced with Zenroom's [special effects](https://dev.zenroom.org/#/pages/lua)
 
 This function accepts a mandatory **SCRIPT** to be executed and some optional parameters:
   * DATA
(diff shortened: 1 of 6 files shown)


docs/pages/javascript.md
```diff
@@ -28,99 +28,145 @@
...
 * * *
 
 ## 🎮 Usage
 
-The binding consists of one main function::
+The binding consists of two main functions:
 
-**zencode_exec** to execute [Zencode](https://dev.zenroom.org/#/pages/zencode-intro?id=smart-contracts-in-human-language). To learn more about zencode syntax look [here](https://dev.zenroom.org/#/pages/zencode-cookbook-intro)
+- **zencode_exec** to execute [Zencode](https://dev.zenroom.org/#/pages/zencode-intro?id=smart-contracts-in-human-language). To learn more about zencode syntax look [here](https://dev.zenroom.org/#/pages/zencode-cookbook-intro)
+- **zenroom_exec** to execute our special flavor of Lua enhanced with Zenroom's [special effects](https://dev.zenroom.org/#/pages/lua)
```
puria, line 60: This was intentionally left out, becasue @jaromil don't really like people using lua... bho... maybe we want to tell about the new introspection, though...

docs/pages/zenroom-javascript2b.md
```diff
@@ -220,8 +220,24 @@ Let's add also the decrypt function and we are ready to deploy our static and su
...
+                }
+              }
+            } else if (l.startsWith('[!]')) {
+              x.push(l);
+              return x;
+            }
+            return x;
+          }, []);
+          rel.value = err.join("\n");
+          rel.classList.add("is-danger");
```
puria, line 240: uff this should really go to the `javascript/bindings` js wrapper library... how do you see her (come la vedi)?

docs/pages/zenroom-javascript3.md
```diff
@@ -12,99 +12,216 @@ Now let's add some interactivity and see how we can play and interact with Zenco
...
+    --js \
+    --src-dir \
+    --disable-git \
+    --app \
+    --use-npm \
+    --no-eslint \
+    --no-tailwind \
+    --no-turbopack \
+    --no-import-alias
+```
```
puria, line 30: I have mixed feelings about this, I know that with next.js everything goes smooth, but I really wanted to avoid to use a framework of a framework... that's why I went with a plain react app... maybe we can do an article on how to use it with different more famous js/ts frameworks also angular/svelte/next/vue, BUT keeping at least one very simple example with plain react...

docs/pages/zenroom-javascript3.md
```diff
@@ -12,99 +12,216 @@ Now let's add some interactivity and see how we can play and interact with Zenco
...
+  webpack: config => {
+    config.resolve.fallback = {
+      fs: false,
+      process: false,
+      path: false,
+      crypto: false
+    };
+    return config;
+  }
+};
```
puria, line 156: bravissimo

docs/pages/javascript.md
matteo-cristino: I saw that we have it on the [npm readme](https://www.npmjs.com/package/zenroom), thus I added it also here.

For the introspection and other APIs I listed them with no example at the end, thinking they were not the main part, but I can expand them with example if we want 🌻

docs/pages/zenroom-javascript2b.md
matteo-cristino: Maybe just a utility API? Like `getErrorFromLogs`? That extract all the strings that start with a `[!]` from the trace and the b64 trace

I would not do it by default since the other error parts are useful in case we want to debug an error

docs/pages/zenroom-javascript3.md
matteo-cristino: ok, I just follow the info in the [react documentation](https://react.dev/learn/start-a-new-react-project). Then will try to use plain react

matteo-cristino: Done with plain react and save the `next` documentation in a different file

docs/pages/zenroom-javascript2b.md
puria: yes something like that, `decode_error` to be consistent with the underscore naming, ideally consumed like this:

```javascript
import {zencode_exec, decode_error} from 'zenroom'


zencode_exec(scripts, {}).then(decode_error);
// same as
const result = await zencode_exec(script, {})
decode_error(result)
```

that does the decode and print them in console.log/error as the normal output <3

PR #975, opened 2024-11-21
docs: update list of all zencode statements

matteo-cristino:
- docs: update script to generate list of zenroom statements
- docs: update list of statements, add missing scenarios and add if and foreach sections
- refactor(ethereum): remove the possibility to use a statement as an if since it has no zencode_assert in it


puria (commented):
Thank you <3 this is awesome, I'm so happy someone is gardening the documentation, I want to send all my support and love! Grazie!!!

PR #982, opened 2024-11-28
feat: add support for array and key as a variable in "copy/move from to" statement

matteo-cristino:
deprecate "create copy of element '' from array ''" in favour of the latter


jaromil:
much better. lets keep in mind this use of from. it somehow matches also its use in the signature schemes.

PR #1001, opened 2025-01-10
fix(random): use unifrom distribution to generate random number with modulo, shuffle arrays and extract subtable

matteo-cristino:
unifrom distribution is obtained by using rejection sampling and reservoir sampling


jaromil:
right on! some notes:

1. the term table is very ambiguous in last sentence
2. the 'of' particle is ambiguous while loaded with meaning (quantity). i think we should use 'long' for array and 'size' for octet size

matteo-cristino:
1. Since last statement can create both array and dictionsries based on what it has in input I was not able to found a better term to reflect that 🤔
2. If I understood it coorectly would be like
```gherkin
When I create random '' bytes size
When I create array '' random long '' bits size
```
to me this format does not seems very correct grammatically speaking

matteo-cristino:
```
create random array with '23' elements
create random array with '23' elements each of '8' bits/bytes
create random array|dictionary with '' elements from ''
```

matteo-cristino:
Following the yesterday idea we should also refactor
* create array of '' random numbers
* create array of '' random numbers modulo ''

to
* create random array with '' numbers
* create random array with '' numbers modulo ''

so that all the statement that create random array have the same format

matteo-cristino:
List of all the deprecated statements and new statement
```
old and deprecated                                      -> new
-------------------------------------------------------------------------------------------------------------------
create random object of '' bits                         -> create random of '' bits
create random object of '' bytes                        -> create random of '' bytes
create array of '' random objects                       -> create random array with '' elements
create array of '' random objects of '' bits            -> create random array with '' elements each of '' bits
create array of '' random objects of '' bytes           -> create random array with '' elements each of '' bytes
create array of '' random numbers                       -> create random array with '' [numbers|integers]
create array of '' random numbers modulo ''             -> create random array with '' [numbers|integers] modulo ''
pick random object in ''                                -> create random pick from ''
create random dictionary with '' random objects from '' -> create random [array|dictionary] with '' elements from ''
```
It looks beautiful ✨

jaromil:
It is really a huge improvement, we can apply the same principles we debated to get here for other statements in the future. Right on!!

PR #1002, opened 2025-01-13
feat: parallel loops over multiple arrays

matteo-cristino:


src/lua/zencode_foreach.lua
```diff
@@ -114,11 +114,55 @@ Foreach("'' in sequence from '' to '' with step ''", function(name, from_name, t
...
+                    n_codec.schema = c.schema
+                    n_codec.zentype = "e"
+                end
+                new_codec(prefixed_name, n_codec)
+            end
+        end
+    end
+end
+
+Foreach("values prefix '' at same position in arrays '' and ''", _zip_with_prefix)
```
matteo-cristino, line 151: values/value

PR #1004, opened 2025-01-14
refactor: create zencode_math.lua file to include all mathematical statements on elements

matteo-cristino:


jaromil:
Excellent! lets keep an eye on utrim() which may belong to zencode_common.lua

also if its used a lot it may be worth an implementation in C

matteo-cristino:
In the explicit math operation statements we have that if a table is passed as input the latter is firstly aggrgated (all values are summed) and the the result is passed as value to the math operation. This can maybe create some confusion among the users, moreover we have explicit statements to aggregate arrays like
```gherkin
When I create aggregation of array ''
When I create sum value of elements in array ''
```
so do we want to keep this feature or we want to deprecate its usage and tell the user to firstly use the aggregation statement and then use math operations on the result of aggregation?

jaromil:
Better to be explicit and not aggregate without the user asking. Zencode should return an error if value is not Computable (number,float or big)

Issue #833, opened 2024-03-14
Keyring should have a fix encoding


matteo-cristino:
Some keys in the keyring follow the CONF encoding during the output phase:

https://github.com/dyne/Zenroom/blob/d08d5162facbf3f6bf8d4bdcee5c4c72fff121cd/src/lua/zencode_keyring.lua#L186-L190

while in the input there is only a fixed encoding (for the above is base64).
I think that generally it is better if the keyring encoding is fixed, so that we are alsways sure with what type of encoding we have to load the key.

jaromil:
In cass we fix the encoding then base64 is preferred as requiring less changes, also is the most interoperable from shell scripts.

PR #1006, opened 2025-01-19
refactor(keyring): replace if statements with table lookup

matteo-cristino:
refactor also file format that now is divided in import/export, zencode_statements and global utilities (a bit easier to navigate)


matteo-cristino:
should I include #833 here?

PR #1007, opened 2025-01-19
feat(js): add decode_error api

matteo-cristino:
- **chore: add yarn to mise tools list**
- **feat(js): add decode_error api**

### DIFF
--- a/.mise.toml
+++ b/.mise.toml
@@ -5,6 +5,7 @@
 [tools]
 emsdk = "4"
 meson = "1"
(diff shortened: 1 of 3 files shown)


matteo-cristino:
add test with emoji!

puria (changes requested):
please double check with a test that holds utf-8 in the string, then it's okay for me <3

puria (approved):
Ahora si!

PR #1014, opened 2025-01-20
feat: add up to date x509 milagro api from miracl/core

jaromil:


jaromil:
Merging half way (missing zencode) as some fixes here are useful in new PRs

PR #1018, opened 2025-01-27
feat(p256): support all form of public keys (raw, compressed and uncompressed)

matteo-cristino:
Add also some p256 primitives as:
* public key to x, y
* public key to compressed form
* public key validation


jaromil:
excellent! merging now so that I can rebase the multikey work which is primarily for p256

PR #1023, opened 2025-01-29
test: add vector tests for p256, eddsa and qp to meson test suite

matteo-cristino:


jaromil:
Whoeps! well spotted @g7240

PR #1024, opened 2025-02-10
Feat/multikey

jaromil:
# Support multikey format decoding

Provide multibase decoding

Provide varint decoding

Provide multikey parsing

Should also leverage some C/lua primitives from #1018


jaromil:
Merging at this stage to support multibase, need to continue adding multikey for specific use-cases in another PR.

PR #1026, opened 2025-02-11
docs: update python documentation

matteo-cristino:


matteo-cristino:
@andrea-dintino @adam-burns do you think that now it is clear how to use zenroom python bindings? See https://github.com/dyne/Zenroom/blob/5bc279d73014ca079ab93d7ab690c86ed4a65ca8/docs/pages/python.md

PR #1027, opened 2025-02-13
test(ed): new test vectors from w3c-vc-di-eddsa

g7240:
The following test vector for EdDSA have been implemented from [w3c specification](https://www.w3.org/TR/vc-di-eddsa/):
- B.1 Representation: eddsa-rdfc-2022
- B.2 Enhanced Example for Representation: eddsa-rdfc-2022
- B.3 Representation: eddsa-jcs-2022
- B.4 Representation: Ed25519Signature2020


matteo-cristino:
this test should be added to `test/vectors/eddsa.bats` file so that it is run with `make check`

PR #1031, opened 2025-02-20
feat: import ML-KEM from mlkem-native/PQ code

FilippoTrotter:
Replace the current ML-KEM implementation with the [mlkem-native implementation from PQCP](https://github.com/pq-code-package/mlkem-native).


FilippoTrotter:
Since we no longer use implementations exclusively from pqclean for pq algorithms, should we rename the folder inside `lib/`?Otherwise, I think it could be misleading.

matteo-cristino:
I do not know if we simply want to change the name of `pqclean` folder or, since now it came from another package, we should create something like `mlkem-native` dir inside `lib` and move it there.

@jaromil what do you think?

BTW I added a README inside the `lib` folder so that we can keep track of where a library comes from, pls once the decision on what to do is taken update it also accordingly 🫶

jaromil:
If it is a different library, then yes it should have its own folder under lib like `lib/mlkem-native`.

jaromil:
See #1032 for a new start at importing lib/mlkem and build with its own system as static lib.

jaromil (commented):
Thanks for all this work, it gets close to a solution.
I've studied it and rewrote #1032 improving a few things about maintainability, we should use that.

PR #1033, opened 2025-02-25
Pool memory manager with sailfish-pool

jaromil:
Here is my new memory manager written and fine tuned precisely for zenroom: [fastalloc32.c](https://github.com/dyne/fastalloc32)

The need for it is growing as the complexity of operations grows and especially post-quantum ones have bigger buffers etc.

 I am following a rather minimalist approach to code changes, overloading *alloc calls from pre-processor and instituting a global ZMM pointer to make the memory manager available to all internal functions, which have just to include <zen_memory.h> and then use malloc/free/realloc as usual withou any needed change.

The institution of a global variable is a step back from reentrancy but its OK since we make no use of it. In fact further optimization will follow as I'll make also  the Z context global and reachable from all internals instead of using casted resolution as Z(L) - this also will need no code changes just pre-processor magic.


jaromil:
Current benchmarks are at par between system and sfpool memory allocation strategies, with the only difference that sfpool gives us more control on the memory manager and zeroes all key-sized memory freed immediately:

```
Zenroom sfpool bench on minis
make check on RELEASE=1 (-O3)
Linux 6.1.0-26-amd64 x86_64 GNU/Linux
12th Gen Intel(R) Core(TM) i9-12900HK

x86 64bit target:

master:
real    0m12.306s
user    2m23.116s
sys     0m24.317s

sfpool:
real    0m12.295s
user    2m24.394s
sys     0m24.893s

WASM 32bit target:

master:
real    0m32.779s
user    7m39.226s
sys     1m1.078s

sfpool:
real    0m33.649s
user    7m55.154s
sys     1m1.679s

```

PR #1034, opened 2025-03-06
Update crypto_bbs.lua

JuliusK9:
Nicola and me updated the bbs.sign function in the crypto_bbs.lua with an additional 'if' condition, which check if the public key is correct and different from nil.


matteo-cristino:
overseeded by #1036

PR #1036, opened 2025-03-06
chore: modified bbs.sign()

nsuzzi:
We have modified bbs.sign:
- adding a check that the public key (pk) is not nil
- modifying the error description 

In addition we have noted, according  with the documentation, that is not necessary to have the public key in the input of bbs.sign(). It should be sufficient to generate it into the function since the public key pk is used only in the core.sign() into bbs.sign().
However if we had removed pk from the input of bbs.sign(), the test should have been modified, too.

### DIFF
--- a/src/lua/crypto_bbs.lua
+++ b/src/lua/crypto_bbs.lua
@@ -381,6 +381,10 @@ OUTPUT: the signature (A,e)
 ]]
 function bbs.sign(ciphersuite, sk, pk, header, messages_octets)
 
+    -- Checking the correctness of the input public key pk 
+    if pk ~= (ECP2.generator() * sk):to_zcash() then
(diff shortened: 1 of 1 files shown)


src/lua/crypto_bbs.lua
```diff
@@ -381,6 +381,10 @@ OUTPUT: the signature (A,e)
 ]]
 function bbs.sign(ciphersuite, sk, pk, header, messages_octets)
 
+    -- Checking the correctness of the input public key pk 
+    if pk == nil then
+            error('Error: public key mismatch',2)
```
puria, line 386: please add a more specific error message eg. refer to BBS

jaromil:
To overcome a regression in commits I've restarted this in #1038 also addressing puria's request

PR #1042, opened 2025-03-11
docs: onepage builder now recognize also fragments

matteo-cristino:
- **docs: onepage builder now recognize also fragments**
- **docs: update list of all the statements**


jaromil:
one step forward! well done thanks!

PR #1047, opened 2025-03-31
fix: remove zeros in arrays

matteo-cristino:


jaromil:
well spotted! needed an update since the new float type

PR #1048, opened 2025-04-01
chore: remove unused table in zero removes

matteo-cristino:
this table was moved to the intern function


jaromil:
merging the when_docs PR, then you can rebase to fix this cmake glitch

PR #1049, opened 2025-04-07
Merkle tree tests

JuliusK9:


src/lua/zencode_merkle.lua
```diff
@@ -0,0 +1,82 @@
...
+
+When("create merkle root of ''", 
+    function(name)
+        local data = ACK[name] 
+        if type(data) ~= 'table' then
+            error("Can only use tables")
+        end
+        ACK.merkle_root = create_merkle_tree(data) 
+        new_codec('merkle root', {zentype = "string"})
+    end
```
matteo-cristino, line 48: Similar functions to this one are written multiple time (one for each statement). We want to avoid code duplication. So this function can be written outside and called each time or it can be integrated directly in create_merkle_tree function.

Bonus point: Improve error message

PR #1052, opened 2025-04-08
feat: new O.from_uuid implemenation

jaromil:


jaromil:
Well done so far! can you please add also the string `urn:uuid:` as a string prefix that is optionally elided in the Given phase? I mean that, if present, the Given statement should not error out, but skip the prefix and read the value following it. If not present, do like it already does.

We will then save this detail in a new CODEC field `prefix` so that the Then statement will restore the prefix string in output.

jaromil:
well done!!

PR #1055, opened 2025-04-24
fix: do not use newlines in deprecation message with compact format

matteo-cristino:


matteo-cristino:
Example of new deprecation message when using compact format:
```
"[W] DEPRECATED: create random object of '' bits, use instead: create random of '' bits"
```

jaromil:
wow thanks this was very annoying

PR #1056, opened 2025-04-24
feat: add extra input to zencode_exec and zenroom_exec api

matteo-cristino:


puria (commented):
Maybe the docs of the JS/TS binding at least should be updated?

jaromil:
Silly question because I don't know JS: is the calling form of the JS/TS function changing, meaning it needs at least 2 more NULL arguments, or it is like in Lua that arguments are omitted and then considered NULL?

jaromil:
I would like this PR to update also _tobuf and add to base zenroom CLI program two flags to pass extra and context.

matteo-cristino:
> I would like this PR to update also _tobuf and add to base zenroom CLI program two flags to pass extra and context.

Will do that ✨

matteo-cristino:
@jaromil I did what you asked:
* extra and context are now present in `zencode_exec`, `zenroom_exec`, `zencode_exec_tobuf` and `zenroom_exec_tobuf` (add also some bats tests for this API)
* zenroom cli now has `e` and `x` option for extra and context and a new help message that should help a bit more the user
* extra and context are also used in JS bindings
* update some docs (not sure if there is more to update)

jaromil:
Excellent! thanks!

PR #1058, opened 2025-04-29
feat: add utility function to write data in path

matteo-cristino:
seems a good start, but it need a lot of testing


jaromil:
can we reuse or modularize the existing path retrieving function? I haven't checked myself right now, lets just use as less code as possible. agree this has an interesting number of corner cases and your seems a very good start

jaromil:
The `set_in_path` internal global function may be more suited for inclusion in `zencode_data.lua` and maybe we should rename it together with the read one as `write_to_path` and `read_from_path`

matteo-cristino:
One thing comes to my mind about this: when finding a number in the path, *e.g.* `table.1`, we treat it like a number and search for it assuming table is an array, but table can also be a dictionary with a string key "1"... To overcome this issue we can use square brackets to inidicate array indexing, so something like `table[1]` will be interpreted as array indexing, while `table.1` as dictionary indexing

PR #1060, opened 2025-05-02
Refactor the SD-JWT implementation to support more signature types

jaromil:
This PR will remove the hard-coded setting of ES256 signature from SD-JWT and enable it to use any signature available in Zenroom, it will also improve ES256 when possible and a few things in default serialization.


jaromil:
moved to #1062 

merging this smol mod for now

PR #1061, opened 2025-05-05
fix: make eddsa zencode sigs codec agnostic

jaromil:
so far eddsa signatures and public keys forced the codec to be base58, this commit changes the behavior to use the default encoding (base64 or anything set by user). This can be a breaking change for some scripts and currently breaks 4 tests.


matteo-cristino:
move to #1067, for the moment warn user for future changes on eddsa encoding. The change will be done in v6

PR #1062, opened 2025-05-05
feat: first mod of sd-jwt to support algo setting in sig

jaromil:
The statement: `create signed selective disclosure of '' with ''` is now added to specify a string with name of algo, which is parsed internally with uppercase strings of algos set in sd-jwt structures

### DIFF
--- a/src/lua/crypto_sd_jwt.lua
+++ b/src/lua/crypto_sd_jwt.lua
@@ -47,6 +47,7 @@ local function export_str_dict(obj)
 		 end
 	  elseif tv == 'zenroom.octet' then res = v:str()
 	  elseif tv == 'boolean' then res = v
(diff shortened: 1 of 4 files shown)


jaromil:
This is summing up quite some changes to W3C code but will make sure that all tests are passed, the only disruptive change so far is the `BEARER ` schema elision.

It is unfinished since it still needs a rewrite of `W3C.create_string_jwk` which does some weird encoding in the WHEN phase rather than staying in binary form, it also needs a refactor to support more crypto algos from the IANA register, at least the 4 ones I'm enabling in this PR.

matteo-cristino:
We have the transformation zencode keyring to jwks, maybe it can also be usefull to have jwks to zencode keyring 🔐

Let me know if it something that you would like to add to this PR or I should make an issue

PR #1063, opened 2025-05-06
feat: fiat-shamir implementation

JuliusK9:


jaromil:
Neat use of octet concatenation as "stream"!

matteo-cristino:
I would avoid the use of uppercase letter in file name and I will follow the convention we are using in general for lua files, thus `libZK_fiat_shamir.lua` can be something like
* `crypto_fiat_shamir.lua`
* `crypto_libzk.lua`

what do you think about it @jaromil ?

jaromil:
> I would avoid the use of uppercase letter in file name and I will follow the convention we are using in general for lua files, thus `libZK_fiat_shamir.lua` can be something like
> 
> * `crypto_fiat_shamir.lua`
> * `crypto_libzk.lua`
> 
> what do you think about it @jaromil ?

yes: `crypto_fiat_shamir.lua` and also include into the source code as comment all source materials used to develop this implementation, including papers, RFCs and vector tests urls.

PR #1065, opened 2025-05-07
fix: update merkle tree

matteo-cristino:
- **added a function for test vectors in merkle_tree.lua, created tests for creation merkle tree check_merkle_tree.lua**
- **chore(style): remove trailing whitespaces**

### DIFF
--- a/src/lua/zencode_merkle.lua
+++ b/src/lua/zencode_merkle.lua
@@ -75,17 +75,16 @@ When("create merkle root of dictionary path '' using hash ''", _zencode_merkle_r
 
 -- Function to verify the integrity of a Merkle root
 local function _verify_merkle_root(root, name)
-    
     local data_table = pick_from_path(name, true)
     local merkle_root = ACK[root]
-    
+
     if not data_table or not merkle_root then
         error("Inserted values should be not nill", 2)
-    end 
+    end
 
     if type(data_table) ~= 'table' then
         error("Table not found in path: "..name, 2)
-    end 
+    end
 
     local computed_root = _create_merkle_root(data_table)
 
@@ -99,7 +98,6 @@ end
 When("verify merkle root '' of ''", _verify_merkle_root)
 When("verify merkle root '' of dictionary path ''", _verify_merkle_root)
 
-
 local function _create_merkle_tree(data_table, hashtype)
     local N = #data_table
     --creation of the empty tree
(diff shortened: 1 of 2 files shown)


jaromil:
very well, there is no fix: commit so I'll wait for merge in case we want to trigger release

matteo-cristino:
TODO:
* add merkle vector test to run with all other test (add it to bats file in vector test)
* vector tests do not work beacuse function called are `local` in zencode_merkletree.lua, maybe create crypto_merkletree.lua as we did for the other packages?
* vector test call a non existing data2 dictioanry

matteo-cristino:
bonus point: maybe _create_merkle_tree can be optimized, I'm not sure if the first for loop where all the tree is set to 0 string is really needed 🔮

src/lua/crypto_merkle.lua
```diff
@@ -0,0 +1,363 @@
...
+        if set[n][2]%2 == 1 then
+            t = _hash(t .. set[n][1])
+        else
+            t = _hash(set[n][1] .. t)
+        end
+        n = n-1
+    end
+    
+    return t 
+    --return t == root
```
matteo-cristino, line 193: should this line be uncommented?

src/lua/crypto_merkle.lua
```diff
@@ -0,0 +1,363 @@
...
+    local m = #indeces_set
+
+    table.sort(indeces_set)
+
+    for i = 1, m do
+        table.insert(set, {proof[i], indeces_set[i]} )
+    end
+
+    local n = #proof    
+    local t = _hash(set[n-1][1] .. set[n][1])
```
matteo-cristino, line 179: hashtype should always be passed to `_hash` function as second argument (here and all in the other functions call)

src/lua/crypto_merkle.lua
```diff
@@ -0,0 +1,363 @@
...
+        table.insert(data, leaf)            -- data not alredy hashed
+        table.insert(leaves, _hash(leaf))   -- leaves have to be hashed
+        table.insert(set, i)                -- creating a set where we will extract random indeces in the next step
+    end
+
+    local tree = MT.create_merkle_tree(data, hashtype)    
+    local leaves_for_proof = {}
+
+    for i = 1, batch_size do
+        random_index = math.random(n+1-i)
```
matteo-cristino, line 349: this should be local

PR #1067, opened 2025-05-09
chore: add warning on eddsa key and signature encoding

matteo-cristino:


jaromil:
Maybe add the word "deprecate" or "deprecation" somewhere even just in a comment, so that a quick search later will include this part to be elided.

jaromil:
awesome!! thanks

PR #1069, opened 2025-05-13
test(vectors): add ecdsa_deterministic test to ci

matteo-cristino:


jaromil:
well spotted! great poetry in this test unit, we should follow example 😃

PR #1070, opened 2025-05-14
ci: run tests with asan

matteo-cristino:


jaromil:
woooh hoo!! this is quite amazing as it makes zenroom more robust now AND later on all PRs. Many thanks for walking this last ASAN miles so well. This is HARDENING

PR #1079, opened 2025-06-16
feat(time): add utc encoding

matteo-cristino:
this can be helpfull for date in w3c vc or did

### DIFF
--- a/src/lua/zencode_data.lua
+++ b/src/lua/zencode_data.lua
@@ -636,7 +636,7 @@ function apply_encoding(src_name, src_enc, dest_enc)
   local encoded_src
   -- accpet also schemas as encoding
   if ZEN.schemas[uscore(src_enc)] then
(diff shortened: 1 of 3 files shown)


matteo-cristino:
using UTC as an encoding is a bit difficult since it is a time standard without timezones or daylight saving and thus it I am not totally sure on how we should encode a timestamp or a date_table to it. Should we start with a statement to simply create UCT timestamp (saved as a string inside zenroom memory)?

jaromil:
> using UTC as an encoding is a bit difficult since it is a time standard without timezones or daylight saving and thus it I am not totally sure on how we should encode a timestamp or a date_table to it. Should we start with a statement to simply create UCT timestamp (saved as a string inside zenroom memory)?

Yes the best would be
```
When I create the UTC timestamp of 'name_of_time_var'
When I create the UTC timestamp of now
```

jaromil (approved):
excellent!

PR #1086, opened 2025-06-23
fix(js): do not parse heap in introspection

matteo-cristino:


bindings/javascript/src/index.ts
```diff
@@ -208,16 +208,12 @@ export const zencode_valid_input = async (
...
+export const introspect = async (zencode: string, props?: ZenroomProps) => {
   try {
     const { result } = await zencode_valid_input(zencode, props);
     return JSON.parse(result).CODEC;
-  } catch ({ logs }) {
-    const heap = JSON.parse(logs)
-      .filter((l) => l.startsWith("J64 HEAP:"))
-      .map((l) => l.replace("J64 HEAP:", "").trim())[0];
-
-    return JSON.parse(Buffer.from(heap, "base64").toString("utf-8"));
```
puria, line 220: The point is that logs is not a JSON, anymore, correct? Let's not parse it as a JSON, but we still can use the "J64 HEAP:" for trimming and have the base64 to do the introspect, let's do that. `introspect`

puria (changes requested):
Well in place of the

bindings/javascript/src/index.ts
jaromil: yes, the base64 now contains a meaningful string that is not a valid JSON but can be printed as-is to provide useful information on HEAP contents.

PR #1111, opened 2025-07-24
fix: avoid double initialization of memory in longfellow hash

matteo-cristino:


jaromil:
Oh! I thought longfellow tests were going to timeout in CI...

PR #1112, opened 2025-07-24
fix: on teardown reset global ZEN pointer to avoid use-after-free

matteo-cristino:


src/zenroom.c
```diff
@@ -368,6 +368,7 @@ void zen_teardown(zenroom_t *ZZ) {
 	ZMM = NULL;
 	free(ZZ);
 	ZZ = NULL;
+	ZEN = NULL;
```
andrea-dintino, line 371: How 1 line of code, can create (or fix) hell :-D

PR #1113, opened 2025-07-24
fix: hide keyring from heap dump

matteo-cristino:


jaromil:
I am looking into this, will try to print only presence and length without contents.

Meanwhile testing also found out that INT is not printed even when 32byte 
image

matteo-cristino:
If you want any help, let me know 🍝

PR #1114, opened 2025-07-25
fix: normalize big import when using BIG.new with numbers and handle better big numbers and strings in input

matteo-cristino:
- **fix: normalize big import when using BIG.new with numbers**
- **fix(zen_big): BIG.new detect when number grater than 32 bits are imported and throw an error and handle also string inputs**


jaromil:
oh that's a relief, to see this fixed

PR #1115, opened 2025-07-25
fix: increase wasm max_mem to 1792MB

matteo-cristino:
till 1536MB wasm was failing running out of memory to generate longfellow circuit


matteo-cristino:
Close since we are not intereseted in creating the circuit on the web... Circuits will be delivered by some trusted authority

PR #1117, opened 2025-07-28
fix(big): when computing invmodp checks that modulo is odd and inputs are coprime

### LINKED ISSUE #856: BIG.modinv works only for prime modules
Looking at the method "modinv" in the file zen_big.c, seems that this method find succesfully the inverse of a big number only if the modulo is a prime number.
When the modulo is a composite number two things can happen:
1. the computation doesn't stop (for example when the big number in input is not invertible).
2. It gives a wrong inverse.

matteo-cristino:
- **feat(milagro): add function to compute gcd between BIGs**
- **fix(big): when computing invmodp checks that modulo is odd and inputs are coprime**


matteo-cristino:
moved to #1187

